PR #90, opened 2024-11-25
use generated cbor error reason in transaction submit testing

ginnun:
# Context
This is far from being ready to review.

- json file format has changed, so i need to update my part
- i need to fix obvious problems

# Important Changes Introduced


michalrus:
Commit bfd64223146799b83eb3c2ce30e11cf2935f0c4f LGTM :+1:

src/cbor/tests/specific.rs
```diff
@@ -0,0 +1,1415 @@
...
+
+#[tokio::test]
+#[allow(non_snake_case)]
+async fn test_cbor_223() {
+    verify_one("81820682820182008306001b000000024f522629820182008201d90102818258205176274bef11d575edd6aa72392aaf993a07f736e70239c1fb22d4b1426b22bc01").await
+}
+
+#[tokio::test]
+#[allow(non_snake_case)]
+async fn test_cbor_222() {
```
michalrus, line 32: Some of these cases are humongous! 🐋 Maybe in the future it makes sense to move them to a separate `test_cbor_222.cbor` file, and only load the contents in this test function?

src/cbor/haskell_display.rs
```diff
@@ -0,0 +1,2651 @@
...
+                } else {
+                    format!("SJust ({})", v.to_haskell_str())
+                }
+            }
+            _ => "SNothing".to_string(),
+        }
+    }
+}
+
+fn is_primitive<T: 'static>() -> bool {
```
michalrus, line 765: I kinda like that – clever ^^

michalrus (approved):
This is a really massive and brilliant amount of work. Very moving. 😳 🥺

I'm also amazed that you were able to pull it off in such time. 💪 🎉 🎊 🚀

All I can say is it Looks Good To Me. ✅

src/cbor/tests/specific.rs
ginnun: @michalrus Yes, I agree.

src/cbor/haskell_display.rs
ginnun: @michalrus Thank you!

Issue #98, opened 2024-12-04
User friendly bootstrap


mmahut:
I would like to provide the SPOs a nice way to bootstrap the cluster. One of the best bootstraps I have seen is the [dolos](https://github.com/txpipe/dolos) project. 

If we can have something like that, basically:

1. Looking for the configuration; either in the local directory or specified by the config flag
2. If not, start the wizard bootstrap, that will ask the user configuration options
3. Generating the configuration file in the local directory

PR #101, opened 2024-12-09
feat: add blockfrost bootstrap #98

gytis-ivaskevicius:
# Context
Implementation of https://github.com/blockfrost/blockfrost-platform/issues/98

# Important Changes Introduced
Flags:
- --init
- --config

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -46,14 +46,14 @@ pub struct RunArgs {
     #[arg(long, default_value = "3000")]
     server_port: u16,
 
-    #[arg(long, required_unless_present_any(&["config"]))]
+    #[arg(long, required_unless_present = "config")]
     network: Option<Network>,
 
     #[serde(default = "LogLevel::default_log_level")]
     #[arg(long, default_value = "info")]
     log_level: LogLevel,
 
-    #[arg(long, required_unless_present_any(&["config"]))]
+    #[arg(long, required_unless_present = "config")]
     node_socket_path: Option<String>,
 
     #[serde(default = "Mode::default_mode")]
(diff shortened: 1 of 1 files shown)


michalrus:
> If someone is aware of some solid layered configuration library

Hey, have you looked at the [`config`](https://docs.rs/config/latest/config/) crate? It [has 2.7k stars](https://github.com/rust-cli/config-rs), and seems like something we need:

> Config organizes hierarchical or layered configurations for Rust applications.
>
> Config lets you set a set of default parameters and then extend them via merging in configuration from a variety of sources:
>
> * Environment variables
> * String literals in well-known formats
> * Another Config instance
> * Files: TOML, JSON, YAML, INI, RON, JSON5 and custom ones defined with Format trait
> * Manual, programmatic override (via a .set method on the Config instance)

The last bullet point could be used for overriding via CLI arguments.

There's also [`figment`](https://docs.rs/figment/latest/figment/), although less popular.

gytis-ivaskevicius:
Yes, I've checked. Both of these libraries do not support CLI arguments nor have integration with Clap

While double-checking this statement somewhere in issue threads I found this library https://github.com/bnjjj/twelf, it seems like what we need tho does not seem to be popular at all. Will give it a shot

michalrus:
But what about using this method from `config`:

> Manual, programmatic override (via a .set method on the Config instance)

Iterating manually over all CLI args shouldn’t be too long, one line per argument?

gytis-ivaskevicius:
how it should work:
- Get CLI arguments
- Get config 
- Merge CLI and config
- Typecheck 

the way you are suggesting would mean that I would need make everything optional, implement checking if its not empty and checking for conflicting flags (like --solitary)

I don't think that that's an option

michalrus:
Can we discuss it a little more, I'm not understanding everything:

> how it should work:
>
> * Get CLI arguments

So that the user can override the location of the config file?

> * Get config
> * Merge CLI and config
> * Typecheck

Do we _have_ to have CLI overrides? The original ticket—#98—doesn’t mention it.

Maybe, for overrides, it's enough to be able to specify multiple config locations and merge them? An advanced user, one that wants to test something quickly, could even do:

```bash
❯ blockfrost-platform run \
    -c ~/.config/default.location \
    -c <(<<<'nodeSocket: /some/where/else.sock')
```

gytis-ivaskevicius:
Closing in favor of https://github.com/blockfrost/blockfrost-platform/pull/131

Issue #83, opened 2024-11-20
Tests: Create basic test for submit transaction


vladimirvolek:
- basic error
- basic success

Issue #82, opened 2024-11-20
Tests: Run integration tests in CI


vladimirvolek:
We have a node ready at https://gitlab.com/blockfrost/blockfrost-platform.

Issue #32, opened 2024-10-31
Make `GET /` more in line with `cardano-cli query tip`


michalrus:
Probably better to keep them consistent between apps.

Here's our current `GET /` response (after #22):

```
❯ curl -fsSL http://0:3000/ | jq --sort-keys .
{
  "name": "blockfrost-platform",
  "version": "0.0.1",
  "healthy": true,
  "nodeInfo": {
    "block": "1cf951872b03e1b588fe4ad47144926d773022277752a95cee8bf6ff9ee8e78d",
    "epoch": 826,
    "era": "6",
    "slot": 71417072,
    "syncProgress": 100.0
  },
  "errors": []
}
```

And `cardano-cli query tip`:

```
❯ cardano-cli query tip --testnet-magic 2 | jq --sort-keys .
{
  "block": 2628189,
  "epoch": 737,
  "era": "Conway",
  "hash": "f157b75012147aec37462a7baff52902bec9f7109e2b04f7a4c8f18cf8b63976",
  "slot": 63719295,
  "slotInEpoch": 42495,
  "slotsToEpochEnd": 43905,
  "syncProgress": "100.00"
}
```

mmahut:
Let's do this before the release if we want to do it, as I hate breaking changes to the outputs once it is released.

gytis-ivaskevicius:
Are we sure about changing it? I kinda like the existing version more

michalrus:
It would only change the `nodeInfo` key:

```diff
--- current.json	2025-01-28 15:06:14.634038845 +0100
+++ cardano-cli.json	2025-01-28 15:03:12.176850660 +0100
@@ -1,7 +1,10 @@
 {
+  "block": 2628189,
   "epoch": 737,
-  "era": "6",
+  "era": "Conway",
-  "block": "f157b75012147aec37462a7baff52902bec9f7109e2b04f7a4c8f18cf8b63976",
+  "hash":  "f157b75012147aec37462a7baff52902bec9f7109e2b04f7a4c8f18cf8b63976",
   "slot": 63719295,
+  "slotInEpoch": 42495,
+  "slotsToEpochEnd": 43905,
-  "syncProgress": 100.0
+  "syncProgress": "100.00"
 }
```

So I created this issue just for consistency, having noticed that they differ, and the users might be used to the `cardano-cli` format.

I think you’re right to like ours more:
* `block` says more than a `hash`.
* Is `"Conway"` better than `"6"`? (Why a string?)
[...]

PR #106, opened 2025-01-13
Integration tests

vladimirvolek:
**PR:**

- Refactored the server code to enable running tests
- Tests now execute in CI (custom runner)
- Added test coverage for the / endpoint and the /tx/submit error
- Cleanup
- Little bit of https://github.com/blockfrost/blockfrost-platform/issues/32
- [x] We need https://github.com/blockfrost/blockfrost-platform/pull/108 first.

resolves: https://github.com/blockfrost/blockfrost-platform/issues/81
resolves: https://github.com/blockfrost/blockfrost-platform/issues/82
resolves: https://github.com/blockfrost/blockfrost-platform/issues/83

related: https://github.com/blockfrost/blockfrost-platform/issues/124


Cargo.toml
```diff
@@ -41,6 +41,7 @@ walkdir = "2.5.0"
 [dev-dependencies]
 rstest = "0.24.0"
 pretty_assertions = "1.4.1"
+lazy_static = "1.5.0"
```
michalrus, line 44: Isn’t `lazy_static` already in Rust’s stdlib? I think I remember I used it somewhere.

Found it – [`std::sync::LazyLock`](https://github.com/blockfrost/blockfrost-platform/blob/961ae46e9d0b78a8b770d5238d36e9fd5429279e/src/cbor/fallback_decoder.rs#L314-L316) in @ginnun's PR.

src/main.rs
```diff
@@ -1,87 +1,42 @@
...
     info!(
-        "Server is listening on {}",
-        format!("http://{}{}/", address, api_prefix)
+        "Server is listening on http://{}:{}/",
+        config.server_address, config.server_port
     );
-    info!("Log level {}", config.log_level);
-    info!("Mode {}", config.mode);
 
+    // 5. Spawn background tasks
```
michalrus, line 35: Why 5. (and 6. below)? Some left over? :)

tests/common.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+pub fn test_config() -> Arc<Config> {
+    let config = Config {
+        server_address: "0.0.0.0".into(),
+        server_port: 8080,
+        log_level: LogLevel::Info.into(),
+        network_magic: 2,
+        mode: Mode::Compact,
+        node_socket_path: "/run/cardano-node/node.socket".into(),
```
michalrus, line 28: Maybe default to that, but allow overriding with an env. variable? E.g. I have it locally in a different location, and had to edit sources to run the tests locally. A variable named `CARDANO_NODE_SOCKET_PATH` is commonly used.

tests/endpoints_test.rs
```diff
@@ -0,0 +1,72 @@
...
+        let body_bytes = to_bytes(response.into_body(), usize::MAX)
+            .await
+            .expect("Failed to read response body");
+        let root_response: RootResponse =
+            serde_json::from_slice(&body_bytes).expect("Response body is not valid JSON");
+
+        assert!(root_response.errors.is_empty());
+        assert_eq!(root_response.name, "blockfrost-platform");
+        assert!(root_response.healthy);
+        assert_eq!(root_response.node_info.sync_progress, 100.0);
```
michalrus, line 38: I started writing that it's sometimes a little less, but you rounded to 2 decimal places above! :bulb:

tests/endpoints_test.rs
```diff
@@ -0,0 +1,72 @@
...
+            .await
+            .expect("Request to /tx/submit failed");
+
+        let cbor = to_bytes(response.into_body(), usize::MAX)
+            .await
+            .expect("Failed to read response body");
+
+        assert_eq!(
+            cbor,
+            "1f2cac5e80a0bdb92cfebe5bad20427b9a8b25b139d3161a2d8bed08328a4596"
```
michalrus, line 69: Wait, but the transaction CBOR is hardcoded above, it won’t work more than once. And here we expect a success and to get a transaction ID back?

I think we need to:
* either expect a failure,
* or generate the transaction each time with [`tx-build` (available in the devshell)](https://github.com/blockfrost/blockfrost-platform/blob/32e5c275e23318dd9b39eb919e5b117705124d0a/nix/devshells.nix#L50-L53) or similar.

michalrus (commented):
Very nice 😊 A couple of questions:

tests/common.rs
```diff
@@ -0,0 +1,44 @@
...
+    tracing_subscriber::fmt::init();
+});
+
+pub fn initialize_logging() {
+    let _ = INIT_LOGGING;
+}
+
+pub fn test_config() -> Arc<Config> {
+    let node_socket_path_env =
+        var("NODE_SOCKET_PATH").unwrap_or_else(|_| "/run/cardano-node/node.socket".into());
```
michalrus, line 23: Maybe `use std::env` + more qualified `env::var("NODE_SOCKET_PATH")`, since `var` is a pretty generic name, and `env::var` is immediately understandable. But whichever you prefer. :)

tests/endpoints_test.rs
```diff
@@ -0,0 +1,86 @@
...
+            .await
+            .expect("Blockfrost request failed");
+
+        let bf_body_bytes = bf_response
+            .bytes()
+            .await
+            .expect("Failed to read Blockfrost response");
+
+        assert_eq!(
+            local_body_bytes, bf_body_bytes,
```
michalrus, line 82: Nice! What would you say to pretty-printing the JSONs first? They should be JSONs almost always. The diff (already quite pretty!) would be easier to read. Something like:

```rust
fn prettify(s: &[u8]) -> String {
    serde_json::to_string_pretty(&serde_json::from_slice::<serde_json::Value>(s).unwrap()).unwrap()
}

assert_eq!(
    prettify(local_body_bytes), prettify(bf_body_bytes),
    …
)
```

michalrus (commented):
I see that `test_submit_route_error` is now [failing to set up Prometheus](https://github.com/blockfrost/blockfrost-platform/actions/runs/12928238337/job/36055330932#step:4:763):

```
---- tests::test_submit_route_error stdout ----
thread 'tests::test_submit_route_error' panicked at src/api/metrics.rs:19:10:
failed to install Prometheus recorder: FailedToSetGlobalRecorder(SetRecorderError { .. })
```

Previously it was happening in `test_root_route`. :thinking:

michalrus (approved):
Great! LGTM :+1: :clap:

Issue #107, opened 2025-01-14
Download `testgen-hs` in a `build.rs` Cargo script


michalrus:
### Plan

- [x] Don’t contact the Internet if a binary is already provided – either on `PATH` or in the current directory (the same logic as when finding it in runtime).
- [ ] If it's been provided, verify the version, and error out telling the user to delete it, and re-run (this will mostly happen in a devshell after version updates).
- [x] If not provided, download the correct version for the correct platform from its GitHub releases.
- [x] Simplify GitHub workflows (tests, and integration tests) – no need to download manually in them.
- [ ] Simplify the `Dockerfile`.
- [x] Check that the Nix build still works.

PR #108, opened 2025-01-14
Build script

vladimirvolek:
Adding a build script to fetch the testgen-hs binary.

Part of: https://github.com/blockfrost/blockfrost-platform/issues/107

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -127,7 +127,7 @@ fn main() {
 }
 
 fn extract_tar_bz2(archive_path: &PathBuf, extract_dir: &PathBuf) {
-    let tar_bz2 = fs::File::open(&archive_path).expect("Failed to open .tar.bz2 archive");
+    let tar_bz2 = fs::File::open(archive_path).expect("Failed to open .tar.bz2 archive");
     let tar = BzDecoder::new(tar_bz2);
     let mut archive = Archive::new(tar);
     archive
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,7 @@
 use bzip2::read::BzDecoder;
 use std::{
-    fs,
+    env,
+    fs::{create_dir_all, write as fs_write, File},
     path::{Path, PathBuf},
     process::Command,
 };
@@ -45,13 +46,16 @@ fn main() {
 
     println!("Fetching {}", file_name);
 
-    // Cache directory for testgen-hs artifacts.
-    let cache_dir = dirs::cache_dir()
-        .expect("Unable to determine cache directory")
+    // Use the project’s target directory instead of a system cache location.
+    let cargo_manifest_dir =
+        env::var("CARGO_MANIFEST_DIR").expect("Unable to find CARGO_MANIFEST_DIR");
+
+    let testgen_dir = PathBuf::from(&cargo_manifest_dir)
+        .join("target")
         .join("testgen-hs")
         .join(testgen_lib_version);
 
-    fs::create_dir_all(&cache_dir).expect("Unable to create cache directory");
+    create_dir_all(&testgen_dir).expect("Unable to create testgen directory");
 
     let archive_name = if target_os == "windows" {
         format!("{}.zip", file_name)
@@ -59,9 +63,9 @@ fn main() {
         format!("{}.tar.bz2", file_name)
     };
 
-    let archive_path = cache_dir.join(archive_name);
+    let archive_path = testgen_dir.join(&archive_name);
 
-    // Download the artifact if not already in the cache.
+    // Download the artifact if not already in the target directory.
     if !archive_path.exists() {
         println!("Downloading from: {}", download_url);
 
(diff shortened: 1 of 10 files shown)


src/cbor/fallback_decoder.rs
```diff
@@ -103,55 +102,6 @@ impl FallbackDecoder {
         }
     }
 
-    /// A heuristic to find the child binary that we’ll use.
-    pub fn locate_child_binary() -> Result<String, String> {
```
michalrus, line 107: I think somethink _like_ that might need to stay here, because we still need to be able to select a correct binary in runtime, when we're not being run by `cargo` (e.g. when using Docker, Nix, or just running the executable directly).

E.g. see this after a `cargo build`:

```
❯ ./target/debug/blockfrost-platform --network preview --node-socket-path ~/.local/share/blockfrost-platform/preview/node.socket --solitary

thread '<unnamed>' panicked at src/cbor/fallback_decoder.rs:119:14:
TESTGEN_HS_PATH environment variable not set. Build script may not have run.: NotPresent
```

src/main.rs
```diff
@@ -29,11 +29,8 @@ async fn main() -> Result<(), AppError> {
     setup_tracing(&config);
 
     // Set up FallbackDecoder
-    info!(
```
michalrus, line 32: Maybe `debug!`? But if not, one can always use `strace` :)

build.rs
```diff
@@ -0,0 +1,159 @@
...
+        testgen_lib_version, file_name, suffix
+    );
+
+    println!("Fetching {}", file_name);
+
+    // Cache directory for testgen-hs artifacts.
+    let cache_dir = dirs::cache_dir()
+        .expect("Unable to determine cache directory")
+        .join("testgen-hs")
+        .join(testgen_lib_version);
```
michalrus, line 52: So the final directory is:

```
~/.cache/testgen-hs/10.1.2.1/extracted/testgen-hs
```

What would you say to maybe adding one more segment, `blockfrost-platform`, i.e.

```
~/.cache/blockfrost-platform/testgen-hs/10.1.2.1/extracted/testgen-hs
```

If anything comes up later, we’ll only ever have a single directory in `~/.cache` :)

-----

OTOH, since this is a _build_ script, I think it should write to the Cargo's `target/` directory of the project? Like where it puts other binaries/libs:

```
❯ cd ~/Work/blockfrost-platform
❯ ls -lh target/debug/deps/libserde_with_macros-1ef82c761f38d70a.so
-rwx------ 1 mw users 3.9M Jan 14 14:23 target/debug/deps/libserde_with_macros-1ef82c761f38d70a.so
```

src/cli.rs
```diff
@@ -86,9 +92,9 @@ pub struct IcebreakersConfig {
 }
 
 impl Config {
-    pub fn from_args(args: Args) -> Self {
+    pub fn from_args(args: Args) -> Result<Self, AppError> {
+        let testgen_hs_path = env::var("TESTGEN_HS_PATH")?;
```
michalrus, line 96: So right now, the user that want to run our executable on bare metal (without Docker, Cargo, Nix) has to remember to export `TESTGEN_HS_PATH` themselves.

If I don’t:

```
❯ ./target/debug/blockfrost-platform --network preview --node-socket-path ~/.local/share/blockfrost-platform/preview/node.socket --solitary

Error: Server("environment variable not found")
```

IMO, as a user, I'd prefer that it was found automatically, what do you think?

build.rs
michalrus: Thanks!

src/main.rs
vladimirvolek: https://github.com/blockfrost/blockfrost-platform/pull/108/files#diff-363108f50c16bc231d3b601409f5c699be655e35a405882de18815aa527a3226R39

src/cbor/fallback_decoder.rs
```diff
@@ -19,21 +22,40 @@ struct FDRequest {
...
+    pub fn spawn() -> Result<Self, AppError> {
+        let mut search_roots = vec![
+            PathBuf::from("/usr/bin"),
+            PathBuf::from("/bin"),
+            PathBuf::from("target/testgen-hs/extracted/testgen-hs"),
+            PathBuf::from("/app/testgen-hs"),
+        ];
+
+        if let Ok(path_var) = env::var("PATH") {
+            for dir in path_var.split(MAIN_SEPARATOR) {
```
michalrus, line 36: There's also [`std::env::split_paths`](https://doc.rust-lang.org/std/env/fn.split_paths.html):

```rust
            for dir in env::split_paths(&path_var) …
```

But either way it works :)

build.rs
```diff
@@ -0,0 +1,187 @@
...
+            "Environment variable {} is set. Exiting the build script.",
+            TESTGEN_HS_PATH
+        );
+        return;
+    }
+
+    // Tell Cargo to rerun the build script if the build script itself changes.
+    println!("cargo:rerun-if-changed=build.rs");
+
+    let testgen_lib_version = "10.1.2.1";
```
michalrus, line 25: This is 10.1.4.0 on `master` for the 10.1.4 `cardano-node`. :)

I wonder, in principle we _could_ read it from the `flake.lock` to define this version in a single place only, something like (Bash):

```
❯ jq <flake.lock -r '.nodes["testgen-hs"].original.ref'
10.1.4.0
```

But maybe it's too much, and it's enough to write a comment in `flake.nix` to also remember to update it here?

build.rs
```diff
@@ -0,0 +1,187 @@
...
+    path::{Path, PathBuf},
+    process::Command,
+};
+use tar::Archive;
+use zip::ZipArchive;
+
+const TESTGEN_HS_PATH: &str = "TESTGEN_HS_PATH";
+
+fn main() {
+    if env::var(TESTGEN_HS_PATH).is_ok() {
```
michalrus, line 14: Now I thought that if the user sets this variable, we don’t verify whether they gave us the correct version. :eyes: But then maybe this verification should be done in runtime, by the `blockfrost-platform` executable?

michalrus (approved):
This turned out to be a lot harder than either of us thought it would be! :hot_face: 

But it works! :muscle: :tada: Thank you. :relieved:

build.rs
michalrus: Moved to:
* #115.

build.rs
michalrus: Moved to:
* #115.

Issue #115, opened 2025-01-17
Versioning of `testgen-hs`


michalrus:
## Plan

- [ ] Verify the version that the user (Nix) gives to the `build.rs` in `TESTGEN_HS_PATH`.
- [ ] Verify the version in runtime, i.e. in `fallback_decoder.rs` ([this `TODO`](https://github.com/blockfrost/blockfrost-platform/blob/3da54717e69c8c9d93f28596e8bb8c48e6e2afa5/src/cbor/fallback_decoder.rs#L124-L128)).
- [ ] Define it in a single place (prob. just read the `flake.lock` in _build_ time). Right now we define it separately in `flake.nix`, `Dockerfile`, and in `build.rs`.
- [ ] Maybe the `Dockerfile` can be simplified? Cargo in there has access to the Internet.

## Follow-up to:

* #108
* #107

vladimirvolek:
@michalrus close?

Issue #104, opened 2025-01-13
Installers (or install scripts) for all platforms


michalrus:
### Plan

- [x] Read what the [Dolos scripts](https://dolos.txpipe.io/quickstart) are doing – it's great UX.
- [ ] Publish our binaries somewhere (RC pre-release, some temp. location?).
- [ ] Release GitHub workflow, which publishes the binaries when triggered.
- [x] Windows – a `setup.exe` would be best.
- [x] Linux – `curl | bash`.
- [x] macOS – a Homebrew tap (see [kupo.rb](https://github.com/CardanoSolutions/homebrew-formulas/blob/main/Formula/kupo.rb)).
- [ ] Also mention Nix packages which are already available.

PR #132, opened 2025-01-27
chore: tweak how `testgen-hs` is found in runtime

michalrus:
While preparing the installers in:
* #104 

… I have noticed that finding the `testgen-hs` helper relied on the _current directory_. It shouldn’t, because the user can run our main executable from any directory, and it should behave in the same way.

Instead, I have added `current_exe_dir`, which points to the parent directory of our executable (`/proc/self/exe` on Linux etc.).

All search paths are also absolute now.

Previously, the Nix build didn’t work – unless the current directory was correct.

I removed directory recursion, as it's not necessary, and not what the kernel usually does when looking at `PATH` – IMO, let's keep it consistent with what the users are used to.

And the same logic is used to check for the file in each directory.

I'm not checking for the version yet, this will be done in:
* #115

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,9 @@ result-*
 
 /cardano-node-configs
 
+testgen-hs
+testgen-hs.exe
+
 # Ignore IDE specific files
 # Visual Studio Code
 .vscode/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -337,7 +337,6 @@ dependencies = [
  "tracing-test",
  "twelf",
  "uuid",
- "walkdir",
  "zip",
 ]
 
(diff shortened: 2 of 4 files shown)


gytis-ivaskevicius (approved):
it looks good, but in the long run we should  get rid of hard coded paths and move env variables into clap so it would be documented

michalrus:
> it looks good, but in the long run we should get rid of hard coded paths and move env variables into clap so it would be documented

I don’t know about this particular case, because this binary is more like a… shared object maybe? Their PATHs are specified in the executable’s header itself (still overrideable with `LD_LIBRARY_PATH`).

E.g. in Postgres you don’t (usually) want to specify locations of helper (libexec) executables – they depend on the location of the main executable, and are relative to it.

I've added a few more cases for convenience, but I don’t think they are _necessary_, e.g. we could (should?) structure the directories in the `Dockerfile` for the first case to just work.

PR #135, opened 2025-01-28
fix: unify json responses

vladimirvolek:
As we already use https://docs.blockfrost.io/#description/errors

```json
{
  "status_code": 403,
  "error": "Forbidden",
  "message": "Invalid project token."
}
```

It would be a good idea to unify JSON responses and consistently use snake_case.


michalrus (approved):
LGTM :+1:

PR #155, opened 2025-02-06
revised introduction

joseph-fajen:
Based on my conversations with Marek, this is my suggested new version of the introduction for review and comments.


mmahut:
@joseph-fajen this is pretty good! I did move the use cases into its own section, to not overload the introduction.

I have also removed the mention of the performance, just as the technical people are going to use it first. :D

mmahut (approved):

PR #157, opened 2025-02-10
feat: parallelize the Haskell tests, and stream the test cases

michalrus:
## Context

* Much less RAM usage – ~300 MB even for the largest generator sizes.

* We also utilize 100% of all CPU cores, and the whole thing takes much less time.

* Related Haskell side changes: [input-output-hk/testgen-hs/compare/10.1.4.1..10.1.4.2](https://github.com/input-output-hk/testgen-hs/compare/10.1.4.1..10.1.4.2).


ginnun (approved):
LGTM!!

PR #171, opened 2025-02-13
chore: check for `bzip2` in `curl-bash-install.sh`

michalrus:
## Context

@mmahut found out that `bzip2` is not installed by default on the latest Fedora right after installation.

Let’s have a better error message than:

```
tar (child): bzip2: Cannot exec: No such file or directory
```


mmahut (approved):

ginnun (approved):
lgtm

choir94:
sudo apt update && sudo apt install bzip2 -y

PR #182, opened 2025-02-18
breaking: Fix metrics endpoint and enable by default

ginnun:
Fixes https://github.com/blockfrost/blockfrost-platform/issues/179

**Breaking Changes**
- Replaced `--metrics` with `--no-metrics` 
- Metrics are enabled by default when running via CLI without  `--no-metrics` 

**Changes**
- `/metrics`  url is no longer handled if metrics are disabled
- HTTP calls are no longer gauged  if metrics are disabled

PR #183, opened 2025-02-19
fix: never let Axum abort Ouroboros miniprotocols on HTTP client disconnect

### LINKED ISSUE #173: N2C connection no longer viable
Causes:
* #175 

## Context

Version 0.0.1 on mac mini, the user is seeing this.

```
INFO N2C connection to node was successfully established at socket: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket
2025-02-14T03:29:02.349443Z ERROR N2C connection no longer viable: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket, BlockfrostError: localstate::ClientError: attempted to send message while agency is theirs, Metrics { created: Instant { tv_sec: 8807, tv_nsec: 756061833 }, recycled: Some(Instant { tv_sec: 9266, tv_nsec: 564873541 }), recycle_count: 165 }
2025-02-14T03:29:42.663635Z  INFO N2C connection to node was successfully established at socket: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket
2025-02-14T03:35:33.141407Z ERROR N2C connection no longer viable: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket, BlockfrostError: localstate::ClientError: attempted to send message while agency is theirs, Metrics { created: Instant { tv_sec: 9311, tv_nsec: 150372625 }, recycled: Some(Instant { tv_sec: 9655, tv_nsec: 812460541 }), recycle_count: 85 }
2025-02-14T03:43:06.270521Z  INFO N2C connection to node was successfully established at socket: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket
2025-02-14T03:43:08.088898Z ERROR N2C connection no longer viable: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket, BlockfrostError: localstate::ClientError: attempted to send message while agency is theirs, Metrics { created: Instant { tv_sec: 9009, tv_nsec: 394637958 }, recycled: Some(Instant { tv_sec: 10113, tv_nsec: 547547375 }), recycle_count: 405 }
2025-02-14T03:44:06.838455Z  INFO N2C connection to node was successfully established at socket: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket
2025-02-14T03:58:21.828983Z  INFO N2C connection to node was successfully established at socket: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket
2025-02-14T03:58:51.287146Z ERROR N2C connection no longer viable: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket, BlockfrostError: localstate::ClientError: attempted to send message while agency is theirs, Metrics { created: Instant { tv_sec: 11030, tv_nsec: 326440166 }, recycled: Some(Instant { tv_sec: 11053, tv_nsec: 957813375 }), recycle_count: 6 }
2025-02-14T04:00:52.538211Z ERROR N2C connection no longer viable: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket, BlockfrostError: localstate::ClientError: attempted to send message while agency is theirs, Metrics { created: Instant { tv_sec: 10114, tv_nsec: 762263833 }, recycled: Some(Instant { tv_sec: 11180, tv_nsec: 239360125 }), recycle_count: 317 }
2025-02-14T04:00:52.632820Z  INFO N2C connection to node was successfully established at socket: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket
2025-02-14T04:04:34.102360Z  INFO N2C connection to node was successfully established at socket: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket
2025-02-14T04:27:40.068507Z ERROR N2C connection no longer viable: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket, BlockfrostError: localstate::ClientError: attempted to send message while agency is theirs, Metrics { created: Instant { tv_sec: 11402, tv_nsec: 602120833 }, recycled: Some(Instant { tv_sec: 12785, tv_nsec: 87122041 }), recycle_count: 351 }
2025-02-14T04:27:40.095534Z  INFO N2C connection to node was successfully established at socket: /Users/vargasmontero/Library/Application Support/Daedalus Mainnet/cardano-node.socket
```

### LINKED ISSUE #175: Too many open files
Caused by:
* #173

## Context

After a while running on my mac mini in oakhost, I get:

Image

The ulimit is small, it is set to 256, but still - we might not be closing some connections down?

michalrus:
Resolves #173

Resolves #175

## Context

Originally reported as:
* #173 

This is a good explanation – https://github.com/blockfrost/blockfrost-platform/issues/173#issuecomment-2666748539:
> * This happens on machines with high load average.
> * Then `GET /` takes a significant while.
> * But the requester doesn’t wait that long, and disconnects.
> * This results in **Axum cancelling the handling future**.
> * And that’s why our code never calls `client.send_release().await` from Pallas.
> * But the connection is returned to the pool in its `impl Drop`.
> * So the next time we try to use it, it’s in incorrect state (not released), and we try to acquire, and that’s an error, which eventually results in `N2C connection no longer viable`.

## Additional work included

I also added a few more gauges to better instrument the code, and understand what’s happening, and I think they will be useful in general, so I left them in.

### DIFF
--- a/src/api/metrics.rs
+++ b/src/api/metrics.rs
@@ -1,4 +1,4 @@
-use crate::{cli::Config, BlockfrostError};
+use crate::BlockfrostError;
 use axum::response::{Extension, IntoResponse};
 use metrics::{describe_counter, describe_gauge, gauge};
 use metrics_exporter_prometheus::{PrometheusBuilder, PrometheusHandle};
@@ -6,22 +6,24 @@ use std::sync::Arc;
 use tokio::sync::RwLock;
 
 pub async fn route(
-    Extension(prometheus_handle): Extension<Arc<RwLock<PrometheusHandle>>>,
-    Extension(config): Extension<Arc<Config>>,
+    Extension(prometheus_handle): Extension<Option<Arc<RwLock<PrometheusHandle>>>>,
 ) -> Result<impl IntoResponse, BlockfrostError> {
-    if !config.metrics {
-        return Err(BlockfrostError::not_found());
+    match prometheus_handle {
+        None => Err(BlockfrostError::not_found()),
+        Some(handle) => {
+            let handle = handle.write().await;
+            Ok(handle.render().into_response())
+        }
     }
-
-    let handle = prometheus_handle.write().await;
-
-    Ok(handle.render().into_response())
 }
 
 pub fn setup_metrics_recorder() -> Arc<RwLock<PrometheusHandle>> {
     let builder = PrometheusBuilder::new()
         .install_recorder()
         .expect("failed to install Prometheus recorder");
+
+    // Note: we’re initializing the gauges with 0, otherwise they’re not present
+    // under `GET /metrics` right after startup, before anything happens.
 
     describe_counter!(
         "http_requests_total",
@@ -32,9 +34,31 @@ pub fn setup_metrics_recorder() -> Arc<RwLock<PrometheusHandle>> {
         "cardano_node_connections",
         "Number of currently open Cardano node N2C connections"
     );
-
-    // Otherwise it’s not present under `GET /metrics` if we start with a failing cardano-node:
     gauge!("cardano_node_connections").set(0);
+
+    describe_gauge!(
+        "cardano_node_connections_initiated",
+        "Number of Cardano node N2C connections that have ever been initiated"
+    );
+    gauge!("cardano_node_connections_initiated").set(0);
+
+    describe_gauge!(
+        "cardano_node_connections_failed",
+        "Number of Cardano node N2C connections that failed and had to be restarted"
+    );
+    gauge!("cardano_node_connections_failed").set(0);
+
+    describe_gauge!(
+        "tx_submit_success",
+        "Number of transactions that were successfully submitted"
+    );
+    gauge!("tx_submit_success").set(0);
+
+    describe_gauge!(
+        "tx_submit_failure",
+        "Number of transactions that were submitted with an error"
+    );
+    gauge!("tx_submit_failure").set(0);
 
     Arc::new(RwLock::new(builder))
 }
--- a/src/api/root.rs
+++ b/src/api/root.rs
@@ -15,8 +15,14 @@ pub async fn route(
     Extension(node): Extension<NodePool>,
 ) -> Result<impl IntoResponse, BlockfrostError> {
     let errors = vec![];
-    let mut node = node.get().await?;
-    let node_info = node.sync_progress().await?;
+
+    // XXX: Axum must not abort Ouroboros protocols in the middle, hence a separate Tokio task:
+    let node_info: NodeInfo = tokio::spawn(async move {
+        let mut node = node.get().await?;
+        node.sync_progress().await
+    })
+    .await
+    .expect("sync_progress panic!")?;
 
     let response = RootResponse {
         name: env!("CARGO_PKG_NAME").to_string(),
--- a/src/api/tx_submit.rs
+++ b/src/api/tx_submit.rs
@@ -1,5 +1,6 @@
 use crate::{common::validate_content_type, BlockfrostError, NodePool};
 use axum::{http::HeaderMap, response::IntoResponse, Extension, Json};
+use metrics::gauge;
 
 pub async fn route(
     Extension(node): Extension<NodePool>,
@@ -9,9 +10,22 @@ pub async fn route(
     // Allow only application/cbor content type
     validate_content_type(&headers, &["application/cbor"])?;
 
-    // Submit transaction
-    let mut node = node.get().await?;
-    let response = node.submit_transaction(body.as_ref()).await?;
+    // XXX: Axum must not abort Ouroboros protocols in the middle, hence a separate Tokio task:
+    let response = tokio::spawn(async move {
+        // Submit transaction
+        let mut node = node.get().await?;
+        let response = node.submit_transaction(body.as_ref()).await;
+
+        if response.is_ok() {
+            gauge!("tx_submit_success").increment(1)
+        } else {
+            gauge!("tx_submit_failure").increment(1)
+        }
+
+        response
+    })
+    .await
+    .expect("submit_transaction panic!")?;
 
     Ok(Json(response))
 }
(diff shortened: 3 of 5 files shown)


src/api/metrics.rs
```diff
@@ -1,28 +1,30 @@
...
 use tokio::sync::RwLock;
 
 pub async fn route(
-    Extension(prometheus_handle): Extension<Arc<RwLock<PrometheusHandle>>>,
-    Extension(config): Extension<Arc<Config>>,
+    Extension(prometheus_handle): Extension<Option<Arc<RwLock<PrometheusHandle>>>>,
 ) -> Result<impl IntoResponse, BlockfrostError> {
-    if !config.metrics {
-        return Err(BlockfrostError::not_found());
+    match prometheus_handle {
```
ginnun, line 11: I have a fix for this, I am working on the tests. This is good for as a quick fix.

src/api/metrics.rs
```diff
@@ -32,9 +34,31 @@ pub fn setup_metrics_recorder() -> Arc<RwLock<PrometheusHandle>> {
         "cardano_node_connections",
         "Number of currently open Cardano node N2C connections"
     );
-
-    // Otherwise it’s not present under `GET /metrics` if we start with a failing cardano-node:
     gauge!("cardano_node_connections").set(0);
 
+    describe_gauge!(
```
ginnun, line 39: Thank you for the new metrics

src/api/root.rs
```diff
@@ -15,8 +15,14 @@ pub async fn route(
     Extension(node): Extension<NodePool>,
 ) -> Result<impl IntoResponse, BlockfrostError> {
     let errors = vec![];
-    let mut node = node.get().await?;
-    let node_info = node.sync_progress().await?;
+
+    // XXX: Axum must not abort Ouroboros protocols in the middle, hence a separate Tokio task:
```
ginnun, line 19: What does `XXX:`  stands for?

ginnun (changes requested):
@michalrus Thank you for the great inspection here! 

I think we can have improvements in the future, some options that are unrelated to eachother:
- non-blocking http and reactive way of handling node communications
- an overseer or manager for those spawned tasks, so they are under control (orphanage etc)
- prevent axum to halt even the remote party cancels

Also, I think we should code or contribute to be safe even somehow  Ouroboros protocols gets aborted in the middle.

src/api/tx_submit.rs
```diff
@@ -9,9 +10,22 @@ pub async fn route(
     // Allow only application/cbor content type
     validate_content_type(&headers, &["application/cbor"])?;
 
-    // Submit transaction
-    let mut node = node.get().await?;
-    let response = node.submit_transaction(body.as_ref()).await?;
+    // XXX: Axum must not abort Ouroboros protocols in the middle, hence a separate Tokio task:
+    let response = tokio::spawn(async move {
```
ginnun, line 14: Humm, just asking: Should we not cancel the submission if remote party halts the request?

src/api/metrics.rs
michalrus: Ah, right! I had to include this to be able to see `GET /metrics` for my investigation, but your fix in:
* #182 

… is better. I tried to keep this one minimal in terms of LOC.

I'll resolve the conflict as soon as your PR gets merged.

src/api/root.rs
michalrus: 😅 I never checked, and always used it to denote something tricky. I.e. as a more important comment than usual comments. As in you have to watch out when changing code around it, or you’ll break it.

According to https://github.com/SAP/styleguides/issues/279#issuecomment-1192321729:

> * Probably popularized by a developer called Bill Joy in 1971, working on the Berkley Software Distribution (BSD)

michalrus:
@ginnun, re:

> we should code or contribute to be safe even somehow Ouroboros protocols gets aborted in the middle

This is really hard to get right. I've been thinking about a few possible solutions. The problem is that:

1. Axum will cancel the handler once the HTTP client disconnects – it will just not run anything on the next `.await`.
    1. This is a good default, we could change it globally by introducing a middleware which spawns everything in a Tokio task, but I would recommend against it.
2. But Ouroboros expects us to run the client release ceremony once we’re done using a particular Ouroboros miniprotocol.
3. Pallas implements this cleanup ceremony as `async fn` – there’s no other way, TBH.
4. So we somehow have to trigger an `async fn` from a synchronous `fn Drop::drop` on the connection wrapper, where we don’t have access to the async runtime.
5. And the other problem is that this cleanup **must be guaranteed to happen before** this connection is given to the next handler from the Deadpool.

We could allow dirty connection states in the pool, and to the cleanup on next `recycle`. That’s one way, but more complicated than wrapping in a `tokio::spawn` which is guaranteed to _not_ be cancelled by Axum.

I have one more optimization of `GET /` in the next PR, so we’re only left with this `tokio::spawn` solution in `POST /tx/submit`.

The future endpoints will query indexer/DB, so they should be good to cancel at any point.

src/api/metrics.rs
ginnun: No worries, whoever gets there first!

ginnun:
> I have one more optimization of `GET /` in the next PR, so we’re only left with this `tokio::spawn` solution in `POST /tx/submit`.
>
I think this should be OK, no one will argue that tx submissions should be cancellable by halting the HTTP connection.
> The future endpoints will query indexer/DB, so they should be good to cancel at any point.

Could you elaborate?

michalrus:
> > The future endpoints will query indexer/DB, so they should be good to cancel at any point.
> 
> Could you elaborate?

Just an example, but I'm suspecting that, say, [`deadpool-postgres`](https://docs.rs/crate/deadpool-postgres/0.14.1) allows you to cancel the asynchronous execution at any point, and brings back the connection to a usable state (cancels any outstanding transactions etc.). The connection to Postgres is less fragile w.r.t. `ROLLBACK TRANSACTION` than our Ouroboros clients.

So it Just Works™, unlike here.

ginnun:
@michalrus In the future, when we have the bandwidth and the will, we could at least keep the spawned tasks in a list/manager, so they are not 'orphans', at least in the name ^^. But this might be nitpicking, so secure to ignore.

ginnun (approved):
PR looks small, but the effort is huge!

vladimirvolek:
Good work!

PR #190, opened 2025-02-20
feat: calculate `sync_progress` in the health monitor

michalrus:
## Context

Do it once (every 2 or 10 seconds), instead of for each request. On machines with high load average, it can take even 15 seconds, as it did on Marek's Cloud Mac Mini in:
* #183 

I also included some house cleaning, but in separate commits for ease of review.


src/api/tx_submit.rs
```diff
@@ -29,3 +32,37 @@ pub async fn route(
...
 }
+
+/// This function allows us to take both hex-encoded and raw bytes. It has
+/// to be a heuristic: if there are input bytes that are not `[0-9a-f]`,
+/// then it must be a binary string. Otherwise, we assume it’s hex encoded.
+///
+/// **Note**: there is a small probability that the user gave us a binary
+/// string that only _looked_ like a hex-encoded one, but it’s rare enough
+/// to ignore it.
+pub fn binary_or_hex_heuristic(xs: &[u8]) -> Vec<u8> {
```
michalrus, line 43: IMO this belongs to the API, closer to the end user.

src/api/tx_submit.rs
```diff
@@ -29,3 +32,37 @@ pub async fn route(
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn proptest_binary_or_hex_heuristic(binary in prop::collection::vec(any::<u8>(), 0..=128)) {
```
michalrus, line 60: And it's a good and simple opportunity to introduce `proptest` (like QuickCheck in Haskell). We already use property testing of `ApplyTxErr`, but we set it up ourselves. This is a popular library for Rust-only tests.

src/cbor/fallback_decoder.rs
```diff
@@ -268,7 +268,10 @@ impl FallbackDecoder {
 
                 // unwrap is safe, the other side would have to drop for a
                 // panic – can’t happen, because we control it:
-                request.response_tx.send(response).unwrap();
+                request
+                    .response_tx
+                    .send(response)
+                    .unwrap_or_else(|_| unreachable!());
```
michalrus, line 274: A little controversial, but IMO `unreachable!()` is cleaner? And conveys the message better. "We’re sure this branch is unreachable, or else kill everything."

src/middlewares/errors.rs
```diff
@@ -30,8 +30,8 @@ pub async fn error_middleware(request: Request, next: Next) -> Result<Response,
         return Ok(BlockfrostError::method_not_allowed().into_response());
     }
 
-    // Transform server errors to internal server error for user
-    if response.status().is_server_error() {
+    // Transform server errors to internal server error for user – except for 503 from the root route
+    if response.status().is_server_error() && response.status() != StatusCode::SERVICE_UNAVAILABLE {
```
michalrus, line 34: Also a little controversial – I'm not sure about this part.

src/health_monitor.rs
```diff
@@ -0,0 +1,64 @@
...
+        errors: vec![],
+        node_info: None,
+    }));
+    let state = state_.clone();
+
+    let state_update_ = Arc::new(Notify::new());
+    let state_update = state_update_.clone();
+
+    tokio::spawn(async move {
+        loop {
```
ginnun, line 37: in the future, once/if we have other background tasks in some numbers, it will make more sense to have a scheduler that will schedule tasks using only one manager thread. 

For the the need, this is the optimal solution :clap:

ginnun (approved):
A nice improvement! lgtm.

PR #198, opened 2025-02-24
chore: add SHA-256 checksums to the Curl-Bash installer

### LINKED ISSUE #145: Add SHA-256 checksums to the `curl-bash-install.sh`
Follow-up to:
- #141

## Context

Right now we’re not testing that the downloaded archive is what we expect.

HTTP doesn’t validate it on its own.

In the recent GitHub outage, we got a broken archive from them (https://github.com/blockfrost/blockfrost-platform/pull/139#pullrequestreview-2584132726), so it can happen.

Then, the error was obvious, and it was impossible to unpack the archive, but it could be more subtle.

michalrus:
Resolves #145

## Context

See #145 for the reasoning.

## Testing

I tested it via [michalrus/bfp-release-testing](https://github.com/michalrus/bfp-release-testing/releases/tag/0.0.1), you can see that this [`curl-bash-install.sh` file there](https://github.com/michalrus/bfp-release-testing/releases/download/0.0.1/curl-bash-install.sh) contains the hashes:

```bash
❯ curl -fsSL https://github.com/michalrus/bfp-release-testing/releases/download/0.0.1/curl-bash-install.sh \
    | grep 'expected_sha256='

  expected_sha256="0b1bcd7d3479859dd927c92dccdba15e41208a14c69c752be5c8c7d621a4daac"
  expected_sha256="d1dad2c70933a03342df9a92150be9c448f8b6c770ddde1381c68690e2f00d94"
  expected_sha256="16a0317df3bb51d55708c5084bb8a335db99486d24061bd33611f3f24573377b"
  expected_sha256="16a0317df3bb51d55708c5084bb8a335db99486d24061bd33611f3f24573377b"
```

The file itself doesn’t work OOTB, because it links to [our real 0.0.1 release](https://github.com/blockfrost/blockfrost-platform/releases/download/0.0.1), and the hashes don’t match it (they’re generated for the Git revision which runs the release).

But I tested it modifying URL to `http://localhost:12345` in a downloaded `curl-bash-install.sh`, and it does install, and work both on Linux, and macOS.

### DIFF
--- a/nix/internal/curl-bash-install.sh
+++ b/nix/internal/curl-bash-install.sh
@@ -14,19 +14,20 @@ kernel="$(uname -s)"
 case "${isa}-${kernel}" in
 "x86_64-Linux")
   target_system="x86_64-linux"
-  ;;
-"aarch64-Linux")
-  target_system="aarch64-linux"
+  expected_sha256="@sha256_x86_64_linux@"
   ;;
 "x86_64-Darwin")
   target_system="x86_64-darwin"
+  expected_sha256="@sha256_x86_64_darwin@"
   ;;
 # Apple Silicon can appear as "arm64-Darwin" rather than "aarch64-Darwin":
 "arm64-Darwin")
   target_system="aarch64-darwin"
+  expected_sha256="@sha256_aarch64_darwin@"
   ;;
 "aarch64-Darwin")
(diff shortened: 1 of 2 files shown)


ginnun (approved):
lgtm

PR #199, opened 2025-02-24
fix: rustfmt config

vladimirvolek:


michalrus (approved):
LGTM!

It does fix those long lines that were not formatted before. :clap: :sparkles:

Issue #196, opened 2025-02-24
Docs link - 404


vladimirvolek:
![Image](https://github.com/user-attachments/assets/9b31433d-b9a6-4733-8f9d-4f9f0520fc19)

https://github.com/blockfrost/blockfrost-platform/docs/src/content/index.mdx

PR #203, opened 2025-02-25
fix: hide toc link

vladimirvolek:
I haven't found a better solution.

resolves: https://github.com/blockfrost/blockfrost-platform/issues/196


michalrus (approved):
LGTM!

Issue #124, opened 2025-01-22
Add more integration tests


vladimirvolek:
- [x] Transaction send success
- [ ] License from icebreakers (no solitary mode)

PR #204, opened 2025-02-26
Integration test for icebreakers registration

vladimirvolek:
resolves: https://github.com/blockfrost/blockfrost-platform/issues/124


ginnun (approved):
lgtm, good luck

tests/endpoints_test.rs
```diff
@@ -139,4 +146,50 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_icebreakers_registrations() -> Result<(), Box<dyn std::error::Error>> {
+        initialize_logging();
+
+        let (app, _, icebreakers_api, api_prefix) = build_app_non_solitary()
+            .await
+            .expect("Failed to build the application");
+
+        let ip_addr: IpAddr = "0.0.0.0".parse().unwrap();
+        let address = SocketAddr::new(ip_addr, 3000);
```
michalrus, line 160: Since these are tests, we could bind to port 0 (i.e. random), and then get the actual value back from the listener (`listener.local_addr().port` or similar).

michalrus (commented):
LGTM! :+1:

tests/endpoints_test.rs
michalrus: Ah, no, wait you _need_ it to be 3000, because that’s what you have open on `iptables`!

PR #206, opened 2025-02-27
Attempted fix for ipv6 issue

frog357:
Attempt to fix the ipv6 issues as reported via discord.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,7 +16,13 @@ async fn main() -> Result<(), AppError> {
     setup_tracing(config.log_level);
 
     let (app, _, icebreakers_api, api_prefix) = build(config.clone().into()).await?;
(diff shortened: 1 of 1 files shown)


michalrus (approved):
LGTM. This is correct, sorry we missed that!

vladimirvolek:
Thanks you!

SmaugPool:
Note that [tokio::net::TcpListener::bind](https://docs.rs/tokio/latest/tokio/net/struct.TcpListener.html#method.bind) uses a [ToSocketAddrs](https://tikv.github.io/doc/tokio/net/trait.ToSocketAddrs.html) trait that can handle automatically IPv4/IPv6 string address conversion if prefered to manual handling.

michalrus:
@SmaugPool yup, we’ll rework this to use `std::net::*` types everywhere, thanks! :)

PR #212, opened 2025-02-28
chore: require `CHANGELOG.md` updates in each PR

michalrus:
## Context

I keep forgetting. 👀


michalrus:
The PR works, but we don’t want that. :x:

PR #215, opened 2025-03-03
chore: check that the chain is advancing in `health_monitor`

### LINKED ISSUE #205: In the health monitor, check that the chain is advancing
## Context

A user might have a `cardano-node` that stopped advancing (for whatever reason, e.g. network misconfiguration, a firewall of their ISP etc.).

@cleverca22’s idea from a different repository:

> at least one thing i can think off immediately:
>
> * check that the chain height is at least advancing, if the **chain hasnt moved in 5 minutes, something is wrong**, and cloudflare shouldnt be routing requests to this backend

Now, we can do it relatively easily (we can have state in the health `loop { }`) after:
* #190

michalrus:
Resolves #205

## Context

In #205.

### DIFF
--- a/src/health_monitor.rs
+++ b/src/health_monitor.rs
@@ -21,6 +21,8 @@ impl HealthMonitor {
     }
 }
 
+const CHAIN_STALE_IF_OLDER_THAN: std::time::Duration = std::time::Duration::from_secs(5 * 60);
+
 pub async fn spawn(node: NodePool) -> HealthMonitor {
     // This initial state is never seen:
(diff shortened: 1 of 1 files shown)


src/health_monitor.rs
```diff
@@ -21,6 +21,8 @@ impl HealthMonitor {
     }
 }
 
+const CHAIN_STALE_IF_OLDER_THAN: std::time::Duration = std::time::Duration::from_secs(5 * 60);
```
michalrus, line 24: Michael suggested 5 minutes (see #205), but maybe we should fail when it’s less than that?

Issue #194, opened 2025-02-21
Add timeouts to `cardano-node` requests over the UNIX socket


michalrus:
## Context

Since Ouroboros protocols are quite formal, the only thing we can do is to scrap the Node2Client connection that timed out.

But this is tricky, as some requests may be expected to take more.

Still, I think we should probably have at least _some_ timeout (make it even 10 min), to be able to self-heal, and keep the connection pool clean in case of weird and unexpected situations (deadlock in the node etc.).

Be less trusting.

PR #227, opened 2025-03-10
Add timeout to NodeClient connection

ginnun:
Resolves https://github.com/blockfrost/blockfrost-platform/issues/194

### DIFF
--- a/src/node/connection.rs
+++ b/src/node/connection.rs
@@ -1,6 +1,7 @@
 use crate::BlockfrostError;
 use pallas_network::{facades::NodeClient as NodeClientFacade, miniprotocols::localstate};
 use std::{boxed::Box, pin::Pin};
+use tokio::time::{Duration, timeout};
 use tracing::error;
 
 /// Our wrapper around [`pallas_network::facades::NodeClient`]. If you only use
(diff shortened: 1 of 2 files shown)


vladimirvolek (approved):

src/node/connection.rs
```diff
@@ -31,8 +36,12 @@ impl NodeClient {
...
             .inspect_err(|e| error!("Failed to acquire a statequery client: {:?}", e))?;
 
-        // Run the action and ensure the client is released afterwards
-        let result = action(client).await;
+        // Run the action with a timeout
+        let result = timeout(duration, action(client)).await.map_err(|_| {
+            let msg = format!("Timeout after {} seconds", duration.as_secs());
+            error!("{} in with_statequery_timeout", msg);
+            BlockfrostError::timeout(msg)
+        })?;
```
michalrus, line 44: This could be just slightly clearer. We have a `Result<Result<A, Err>, tokio::Elapsed>`, and we apply the `?` only on the top level result, because then such a timed out connection is useless in the future, as it’s probably in an incoherent state.

Hmmm… maybe a comment? 👀

michalrus (commented):
LGTM!

PR #236, opened 2025-03-17
Handle invalid tx cbors nicer in submit

ginnun:
Please note that this errors are not word to word identical to the Haskell implementation. In my opinion It should be fine like this.

The specific different error message comes from the language's implementation of serializers.

Haskell:
`DecoderErrorDeserialiseFailure \"Shelley Tx\" (DeserialiseFailure 0 (\"expected list len or indef\"))`

Rust:
`DecoderErrorDeserialiseFailure \"Shelley Tx\" (DeserialiseFailure 0 (\"unexpected type map at position 0: expected array\"))`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2188,7 +2188,7 @@ checksum = "b15813163c1d831bf4a13c3610c05c0d03b39feb07f7e09fa234dac9b15aaf39"
 [[package]]
 name = "pallas"
 version = "0.32.0"
-source = "git+https://github.com/blockfrost/pallas.git?rev=35990b2a1dcecf8d45801a9bb8a3514fda1482eb#35990b2a1dcecf8d45801a9bb8a3514fda1482eb"
+source = "git+https://github.com/blockfrost/pallas.git?rev=35c7b4049d22cdeb866d9f208a70622c1b41c16a#35c7b4049d22cdeb866d9f208a70622c1b41c16a"
 dependencies = [
  "pallas-addresses",
  "pallas-codec",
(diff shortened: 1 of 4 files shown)


michalrus (approved):
> not word to word identical to the Haskell implementation. In my opinion It should be fine like this.

I agree, let’s see if others do too 👀

This error message is different for very wrong "transactions", right? So it should be fine.

vladimirvolek:
Unfortunately, we need the same response everything else is a breaking change for our customers. :/

ginnun:
Hmm, that would be unfortunate. :crying_cat_face: 

But the difference is not going to be a breaking change in my honest opinion. The format of the message will be identical, only the decoding error message will differ. And this decoding error comes from external libraries, for Rust it's minicbor,

ginnun:
Hmm, that would be unfortunate. :crying_cat_face: 

The difference is not going to be a breaking change in my honest opinion. The format of the message will be identical, only the decoding error message will differ. And this decoding error comes from external libraries, Rust's `minicbor`,  Haskell's `cborg`.

My concern is that, there is no obvious way like using the `testgen-hs` and populating all the cbor errors for me to tackle.

I can of course handle the most obvious ones (like empty cbor, or "expected list len or indef") and fallback to the Rust's message for others. 

What do you think?

michalrus:
@ginnun what triggers this particular error?

Is it a "transaction" with a CBOR that doesn't even deserialize to _any_ transaction, regardless of the correctness of said transaction? So it's a _serialization_ error on the part of the user, not a _logical_ error?

If so, I would assume that most users create transactions with some library, so they would almost never run into this. Unless they're constructing them manually?

vladimirvolek:
I would say let's merge this because it's better and maybe it will be enough. But I remember some users even parsing the error message and creating logic around it...

michalrus:
> I would say let's merge this because it's better and maybe it will be enough. But I remember some users even parsing the error message and creating logic around it...

😱

PR #248, opened 2025-03-19
Update pallas and change some logging levels

ginnun:
1. Update pallas commit hash to prevent i32 overflow
2. Promote some log levels so user reports will contain more info

### DIFF
--- a/src/node/transactions.rs
+++ b/src/node/transactions.rs
@@ -7,7 +7,7 @@ use pallas_network::miniprotocols::{
     localtxsubmission::{EraTx, Response},
 };
 use pallas_primitives::conway::Tx;
-use tracing::{debug, info};
+use tracing::{error, info};
 
 impl NodeClient {
     /// Submits a transaction to the connected Cardano node.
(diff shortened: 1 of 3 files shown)


src/node/transactions.rs
```diff
@@ -46,14 +46,28 @@ impl NodeClient {
...
             },
             Err(e) => {
-                let error_message = format!("Error during transaction submission: {:?}", e);
+                let error_message = format!(
+                    "Error during transaction submission: {:?}, CBOR: {}",
+                    e,
+                    hex::encode(&tx)
+                );
                 self.invalidate_connection(&error_message); // Never use this connection again.
+                error!(
```
michalrus, line 64: I _think_ (haven’t checked) that you'll get 3 error messages for the same event:
1. From `self.invalidate_connection`
2. From this `error!`
3. From the error middleware for the `BlockfrostError`

Maybe it’s fine for now

michalrus (approved):
LGTM, thanks!

src/node/transactions.rs
ginnun: Yes, we are already getting twice of each error :(.
We will need to tidy up the logs in the future.
But for now, over-logging is better than under-logging in my opinion.

PR #250, opened 2025-03-20
chore: expose process metrics (memory, CPU time, fds, threads)

michalrus:
## Context

* Some SPOs already define their own Grafana dashboards.

* Also in the past we had problems with number of open FDs.

* This is cheap, and won’t hurt to have.

* The 2nd commit is just Nix noise to expose `libclang.dylib` to the builds on macOS’es.

## Result

```
❯ curl -fsSL http://0:3000/metrics | grep -E '^[a-z]' | sort

cardano_node_connections 1
cardano_node_connections_failed 0
cardano_node_connections_initiated 1
http_requests_total{method="GET",path="/metrics",status="200"} 31
process_cpu_seconds_total 0
process_max_fds 1024
process_open_fds 12
process_resident_memory_bytes 19488768
process_start_time_seconds 1742464582
process_threads 17
process_virtual_memory_bytes 1139335168
process_virtual_memory_max_bytes 0
tx_submit_failure 0
tx_submit_success 0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Added
 
+- Expose process metrics (memory, CPU time, fds, threads)
 - Add more logs and finer details to `NodeClient::submit_transaction`
 - Header `blockfrost-platform-response` in `tx_submit` endpoint
 - Add `aarch64-linux` builds to release artifacts and installers.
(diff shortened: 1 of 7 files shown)


ginnun (approved):
lgtm

PR #253, opened 2025-03-20
chore: tweak installation docs (we’re past a proper release)

michalrus:
## Context

After the 0.0.2 release today, I looked at https://platform.blockfrost.io/ and found a few issues.


docs/src/content/installation/source.mdx
```diff
@@ -37,8 +37,7 @@ cd blockfrost-platform
 git checkout main
 
 # To build a release version (recommended)
-# NOTE: this option will be available after the first release
-# git checkout 0.0.2
+git checkout 0.0.2
```
michalrus, line 40: I'm not sure whether this was commented out because:
1. the tag didn’t exist, or
2. we want to keep this copy-paste’able, and should only have a single `git checkout` command? So now I should comment out the line with `git checkout main`? 🤔

joseph-fajen: I also don't know why this was commented out. Seems like it would be best to update the page so that it speaks to the present circumstances, updating in the future when needed.

joseph-fajen (approved):
Looks good to me.

PR #221, opened 2025-03-05
fix: expose `/` and `/metrics` without the UUID prefix

michalrus:
Resolves #220

## Context

In #220.

Additionally, I have also made the random prefix ignored in metrics.

We don’t currently use `MatchedPath`, which always matches the full prefix of a nested route. But we’ll need something better once we do.

Issue #220, opened 2025-03-05
/metrics should be at the root of the path, not after the random UUID


SmaugPool:
Currently (`blockfrost-platform 0.0.1 (4a906c1ebd3746c3c69f0520e18257feeecde1e2)`), `/metrics` when enabled are only accessible after the random UUID:
```
$ curl -v http://0.0.0.0:3000/4ecb50d8-aff2-4c9e-8064-5f028969407c/metrics
*   Trying 0.0.0.0:3000...
* Connected to 0.0.0.0 (127.0.0.1) port 3000
> GET /4ecb50d8-aff2-4c9e-8064-5f028969407c/metrics HTTP/1.1
> Host: 0.0.0.0:3000
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< content-type: text/plain; charset=utf-8
< content-length: 1011
< date: Wed, 05 Mar 2025 08:41:13 GMT
< 
# HELP http_requests_total HTTP calls made to blockfrost-platform API
# TYPE http_requests_total counter
http_requests_total{method="GET",path="/4ecb50d8-aff2-4c9e-8064-5f028969407c",status="200"} 13
...
```
As the path is not predictable, this prevents to use monitoring from a remote prometheus.

`/metrics` should likely be instead at the root of the path but this currently returns 404:
```
$ curl -v   http://0.0.0.0:3000/metrics
*   Trying 0.0.0.0:3000...
* Connected to 0.0.0.0 (127.0.0.1) port 3000
> GET /metrics HTTP/1.1
> Host: 0.0.0.0:3000
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 404 Not Found
```

michalrus:
I would argue that the health endpoint, `GET /`, should be, too.

ginnun:
@michalrus I think what said above is also valid for the health endpoint which is in our case the root `/`

PR #258, opened 2025-03-25
fix: treat UUID as an implementation detail in docs

michalrus:
## Context

Follow-up to:
* #220 
* #221

@SmaugPool astutely [points out](https://discord.com/channels/826816523368005654/1283801391256698911/1353990968034070598) that we no longer need to rely on UUIDs in documentation.

That’s true, let’s get rid of them.


joseph-fajen (approved):
Looks good to me. Glad to know about this update.

PR #263, opened 2025-03-28
feat: use a WebSocket load balancer to allow connections from behind NAT(s)

### LINKED ISSUE #178: Allow IceBreakers connection from behind NAT(s), and handle public IP changes
## Problem

1. My home server’s public IP (where I run my own `blockfrost-platform` on mainnet) changes each night, because I have a script that selects the best VPN server at the time. This is not being detected in any way, so the requests just stop being routed there, even though the platform is still running.

2. Similar situation: right now we require a public routable IP address (or at least a single port), but some people – especially the ones running Daedalus – will be behind a NAT (e.g. I’m behind 3 NATs, but I have 5 public ports from my VPN provider, which I configured for `blockfrost-platform` and `cardano-node`). We don’t support that currently.

3. Currently, the traffic between Blockfrost.io and `blockfrost-platform` is not encrypted, so eavesdropping and man-in-the-middle is possible.

4. Some SPOs also understandably want to limit firewall input rules as much as possible. After this, a simple `conntrack` rule will be enough (which most people already have).

## A solution

Since [Blockfrost.io](http://blockfrost.io/) is routing everything either way, it would make `blockfrost-platform` much more accessible, even from the deepest networks, if _it_ connected to [blockfrost.io](http://blockfrost.io/), instead of the other way around, like it is now.

It’s similar to how [Cloudflare Tunnel, `cloudflared` (formerly Argo Tunnel)](https://github.com/cloudflare/cloudflared) works. You set it up on your server behind multiple NATs, and it connects to Cloudflare, which is able to then proxy content from your hidden server, which doesn’t need to allow any INPUT rules except for already established connections.

E.g. we could have a very simple REST-over-WebSocket protocol, where:
* `blockfrost-platform` connects to IceBreakers (solves 2.),
* and there's a periodic `keep-alive` message sent by both sides (solves 1.).

The original HTTP API would still listen locally for local usage with `--solitary` or even without it.

I'm proposing to reverse the arrows on the right:

Image

michalrus:
Resolves #178

## Context

- #178

- The server-side PR (`8 files changed, 1220 insertions(+), 71 deletions(-)`) is in the second repository.

## The new (additional) flow

1. In non-solitary mode, when registering with the `blockfrost-icebreakers-api`, the `blockfrost-platform` gets an additional field in the successful response, with a list of load balancers to connect to:

   ```json
   {
     "route": "b0be51e5-e5fe-4e42-98a9-2246173bea19",
     "load_balancers": [
       {
         "uri": "ws://127.0.0.1:3001/ws",
         "access_token": "QNo3SYA0iPY+LfEhb2Gi5NaUOXqLw4b7ZQyg9DL4hFs="
       }
     ]
   }
   ```

2. Currently, this new WebSocket endpoint is served by the same IceBreakers API, because we already have it set up. So it’s easiest to run it there for the initial testing. But this part can easily be extracted, and run elsewhere, perhaps on more than one machine for redundancy.

3. `blockfrost-platform` then connects to the WebSocket endpoint, and sets `Authorization: Bearer <access_token>`.

4. Both sides exchange `Ping` and `Pong` messages continuously, which allows us to detect the other side going offline quite quickly, even with a `-j DROP`. And also to measure the pure network round-trip time, without the overhead of any advanced calculations.

5. The load balancer code exposes a `GET /stats` endpoint, where you can see all currently connected IceBreakers:

   ```json
   ❯ curl -fsSL http://icebreakers_api:3001/stats | jq .
   {
     "IcebreakerX": {
       "api_prefix": "b0be51e5-e5fe-4e42-98a9-2246173bea19",
       "network_rtt_seconds": 0.000387182,
       "connected_since": "2025-03-28T19:09:46.399124184Z",
       "requests_sent": 0,
       "responses_received": 0,
       "requests_in_progress": 0
     }
   }
   ```

6. But a more important new route in the load balancer (currently running as part of the IceBreakers API) is `GET /{api_prefix}/{path}`, and `POST /{api_prefix}/{path}`:

   1. Using the example above, once you issue, say: `curl -fsSL https://icebreakers-api.blockfrost.io/b0be51e5-e5fe-4e42-98a9-2246173bea19/`,
   2. this request will be translated into a JSON message, and sent over the WebSocket connection with IcebreakerX (identified by its `api_prefix`):

      ```json
      {
        "id": "04810b96-87cd-4af0-8c1b-7b5e00806769",
        "method": "GET",
        "path": "/",
        "header": [
          {
            "name": "host",
            "value": "0.0.0.0:3001"
          },
          {
            "name": "user-agent",
            "value": "curl/8.11.1"
          },
          {
            "name": "accept",
            "value": "*/*"
          }
        ],
        "body_base64": ""
      }
      ```
   3. Then this particular `blockfrost-platform` will resolve the request against its local `axum::Router` – all cheap, and in memory, without opening any new TCP connections to itself.
   4. The response is encoded back into a JSON response message, and sent over the WebSocket to the load balancer:

      ```json
      {
        "id": "04810b96-87cd-4af0-8c1b-7b5e00806769",
        "code": 200,
        "header": [
          {
            "name": "content-type",
            "value": "application/json"
          },
          {
            "name": "content-length",
            "value": "241"
          }
        ],
        "body_base64": "eyJuYW1lIjoiYmxvY2tmcm9zdC1wbGF0Zm9ybSIsInZlcnNpb24iOiIwLjAuMiIsInJldmlzaW9uIjoiZGlydHkiLCJoZWFsdGh5Ijp0cnVlLCJub2RlX2luZm8iOnsiYmxvY2siOiIyNjkzZTk1M2M4NmU0NzZlNGZhMmFhYWY2MDZjZWRjOThhZmU1ZjEwZTFhZDViYTk3NmNmZDkxNzI4NGI4MGJiIiwiZXBvY2giOjg4NSwiZXJhIjo2LCJzbG90Ijo3NjU0MjI5Niwic3luY19wcm9ncmVzcyI6OTkuOTV9LCJlcnJvcnMiOltdfQ=="
      }
      ```
   5. Which in turn converts it to a `hyper::Response`, and sends it to the end user.
   6. All `GET` and `POST` requests will work. It’s been tested to successfully submit a new transaction.

   Example:

   ```json
   ❯ curl -fsSL http://icebreakers_api:3001/b0be51e5-e5fe-4e42-98a9-2246173bea19/ | jq .
   {
     "name": "blockfrost-platform",
     "version": "0.0.2",
     "revision": "dirty",
     "healthy": true,
     "node_info": {
       "block": "6d87f87a0642fd4a59bb8af89400e8c7a777e290d2c2e02ceb93059ac7b4aaba",
       "epoch": 885,
       "era": 6,
       "slot": 76533840,
       "sync_progress": 100.0
     },
     "errors": []
   }
   ```

7. The “old” `blockfrost-platform` API remains as-is on port 3000. After the WebSocket testing period it will no longer need to be open to the public, or have a UUID `api_prefix`.

8. For the time being, by using the same UUID v4 prefix structure to communicate with a particular `blockfrost-platform`, it’s enough to just replace its IP address part of the URL with https://icebreakers-api.blockfrost.io – keeping the request path, body, headers etc. the same.

### DIFF
--- a/src/icebreakers_api.rs
+++ b/src/icebreakers_api.rs
@@ -138,6 +138,11 @@ impl IcebreakersAPI {
                         .flatten()
                         .map(|lb| {
                             if lb.uri.starts_with("//") {
+                                info!(
+                                    "load balancer: falling back to {} for a schemeless load balancer URI: {}",
+                                    fallback_proto,
+                                    lb.uri
+                                );
                                 LoadBalancerConfig {
                                     uri: format!("{}{}", fallback_proto, lb.uri),
                                     ..lb
--- a/src/icebreakers_api.rs
+++ b/src/icebreakers_api.rs
@@ -1,6 +1,8 @@
 use crate::{
     cli::{Config, Network},
     errors::AppError,
+    load_balancer::LoadBalancerConfig,
+    server::ApiPrefix,
 };
 use reqwest::Client;
 use serde::{Deserialize, Serialize};
@@ -16,7 +18,7 @@ pub struct IcebreakersAPI {
     mode: String,
     port: u16,
     reward_address: String,
-    api_prefix: String,
+    api_prefix: ApiPrefix,
 }
 
 #[derive(Deserialize)]
@@ -28,11 +30,26 @@ struct ErrorResponse {
 #[derive(Deserialize, Serialize, Debug)]
 pub struct SuccessResponse {
     pub route: String,
+    /// Experimental: a list of WebSocket URIs and access tokens that the
+    /// `blockfrost-platform` should connect to. Blockfrost.io request and
+    /// responses, as well as network reconfiguration requests (in the future)
+    /// will be will be passed to the `blockfrost-platform` over the socket(s),
+    /// eventually eliminating the need for each relay to expose a public
+    /// routable port, and making network configuration on their side much
+    /// easier. We keep the previous setup and backwards compatibility, and just
+    /// observe this experiment.
+    ///
+    /// It has to temporarily be an option, to keep compatibility with the older
+    /// IceBreakers API.
+    pub load_balancers: Option<Vec<LoadBalancerConfig>>,
 }
 
 impl IcebreakersAPI {
     /// Creates a new `IcebreakersAPI` instance or logs a warning if not configured
-    pub async fn new(config: &Config, api_prefix: String) -> Result<Option<Arc<Self>>, AppError> {
+    pub async fn new(
+        config: &Config,
+        api_prefix: ApiPrefix,
+    ) -> Result<Option<Arc<Self>>, AppError> {
         let api_url = match config.network {
             Network::Preprod | Network::Preview => "https://api-dev.icebreakers.blockfrost.io",
             Network::Mainnet => "https://icebreakers-api.blockfrost.io",
@@ -88,7 +105,7 @@ impl IcebreakersAPI {
             "mode": self.mode,
             "port": self.port,
             "reward_address": self.reward_address,
-            "api_prefix": self.api_prefix.strip_prefix("/"),
+            "api_prefix": self.api_prefix.0.unwrap_or_default(),
         });
 
         let response = self
@@ -105,6 +122,34 @@ impl IcebreakersAPI {
             })?;
 
             info!("Successfully registered with Icebreakers API.");
+
+            // In case we get a URI without a protocol (http: or https: or ws: or wss:):
+            let fallback_proto = if self.base_url.starts_with("https:") {
+                "wss:"
+            } else {
+                "ws:"
+            };
+
+            let success_response = SuccessResponse {
+                load_balancers: Some(
+                    success_response
+                        .load_balancers
+                        .into_iter()
+                        .flatten()
+                        .map(|lb| {
+                            if lb.uri.starts_with("//") {
+                                LoadBalancerConfig {
+                                    uri: format!("{}{}", fallback_proto, lb.uri),
+                                    ..lb
+                                }
+                            } else {
+                                lb
+                            }
+                        })
+                        .collect(),
+                ),
+                ..success_response
+            };
 
             Ok(success_response)
         } else {
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,8 +1,10 @@
-use axum::{ServiceExt, extract::Request};
-use blockfrost_platform::{cli::Args, errors::AppError, logging::setup_tracing, server::build};
+use blockfrost_platform::{
+    cli::Args, errors::AppError, load_balancer, logging::setup_tracing, server::build,
+};
 use dotenvy::dotenv;
-use tokio::{signal, sync::oneshot};
-use tracing::info;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+use tracing::{error, info};
 
 #[tokio::main]
 async fn main() -> Result<(), AppError> {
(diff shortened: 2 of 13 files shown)


michalrus:
I used this diff to test against a local `blockfrost-icebreakers-api`, it will probably be useful to reviewers:

```diff
diff --git a/src/icebreakers_api.rs b/src/icebreakers_api.rs
index d28d554..85ce72a 100644
--- a/src/icebreakers_api.rs
+++ b/src/icebreakers_api.rs
@@ -51,8 +51,9 @@ impl IcebreakersAPI {
         api_prefix: ApiPrefix,
     ) -> Result<Option<Arc<Self>>, AppError> {
         let api_url = match config.network {
-            Network::Preprod | Network::Preview => "https://api-dev.icebreakers.blockfrost.io",
-            Network::Mainnet => "https://icebreakers-api.blockfrost.io",
+            // Network::Preprod | Network::Preview => "https://api-dev.icebreakers.blockfrost.io",
+            // Network::Mainnet => "https://icebreakers-api.blockfrost.io",
+            _ => "http://127.0.0.1:3001",
         };

         match &config.icebreakers_config {
```

src/icebreakers_api.rs
```diff
@@ -106,6 +123,34 @@ impl IcebreakersAPI {
...
+            let success_response = SuccessResponse {
+                load_balancers: Some(
+                    success_response
+                        .load_balancers
+                        .into_iter()
+                        .flatten()
+                        .map(|lb| {
+                            if lb.uri.starts_with("//") {
+                                LoadBalancerConfig {
+                                    uri: format!("{}{}", fallback_proto, lb.uri),
```
ginnun, line 142: Optional: It would be nice to log that we are using a fallback protocol to the user.

src/icebreakers_api.rs
```diff
@@ -106,6 +123,34 @@ impl IcebreakersAPI {
...
             info!("Successfully registered with Icebreakers API.");
 
+            // In case we get a URI without a protocol (http: or https: or ws: or wss:):
+            let fallback_proto = if self.base_url.starts_with("https:") {
+                "wss:"
+            } else {
+                "ws:"
+            };
+
+            let success_response = SuccessResponse {
```
ginnun, line 133: Optional: It would be nice to log the load balancer information we got. It might help with troubleshooting once needde.

src/main.rs
```diff
@@ -22,36 +24,77 @@ async fn main() -> Result<(), AppError> {
...
             // Notify that the server has reached the listening stage
-            let _ = ready_tx.send(());
+            notify_server_ready.notify_one();
 
             server_future.await
         }
     });
 
-    if let Ok(()) = ready_rx.await {
-        info!("Server is listening on http://{}{}", address, api_prefix);
```
ginnun, line 50: Optional: We might want to also mention a log about ws

ginnun (changes requested):
Great work!
All my inline comments are optional, so feel free to just ignore them.

On top of them, I think it would great if:

1. Add change log, since this is a big improvement.
2. Mention WS and the diagram to the docs

src/icebreakers_api.rs
michalrus: Hmmm, but with the `access_token`? 👀

The URI will be logged soon after this:

https://github.com/blockfrost/blockfrost-platform/blob/2f4cab53186e39c3db6d775777d80a14658011ac/src/load_balancer.rs#L273

src/main.rs
michalrus: Similar response:

The URI(s) of load balancer(s) will be logged soon after this:

https://github.com/blockfrost/blockfrost-platform/blob/2f4cab53186e39c3db6d775777d80a14658011ac/src/load_balancer.rs#L273

src/icebreakers_api.rs
michalrus: Fixed in 96ab45bf4809b21a89409c6b7319927327de56c3.

michalrus:
@ginnun:
* changelog added in c21d2d93a2da14174ed4a8ba247c75c8971ce507 – thanks!
* I have to think about the docs, what to use, Mermaid?

```mermaid
graph TD;
    A-->B;
    A-->C;
    B-->D;
    C-->D;
```

ginnun:
> @ginnun:
> 
> * changelog added in [c21d2d9](https://github.com/blockfrost/blockfrost-platform/commit/c21d2d93a2da14174ed4a8ba247c75c8971ce507) – thanks!
> * I have to think about the docs, what to use, Mermaid?
> 
> ```mermaid
> graph TD;
>     A-->B;
>     A-->C;
>     B-->D;
>     C-->D;
> ```

Thanks. I have no preference over the docs. I would keep it simple.

ginnun (approved):
Great work!
Docs are optional, can can be finished later.

michalrus:
@ginnun Yeah, let me do the docs in a separate PR, and then also ask @joseph-fajen for a review.

PR #292, opened 2025-04-07
Correct pallas commit hash

ginnun:
Replace existing pallas commit  hash (which doesn't exists) with an existing one.


michalrus (approved):
Thanks! This previous non-existing commit broke the Nix build on machines that didn’t manage to cache it before. I wonder what happened!

Here’s the diff of this change – [pallas/compare/f74c49e40..1dad01ca1](https://github.com/blockfrost/pallas/compare/f74c49e409570e7835a424bd20ad0cf9e86e0efb..1dad01ca12367a25343e8afdd9230e83927f8ed8):

> 53 changed files with 1,321 additions and 3,549 deletions

Issue #181, opened 2025-02-18
unit test configurations


ginnun:
I discovered that metrics are always on when used via the command line with or without `--metrics`.

Tests can catch these kinds of simple bugs.

PR #302, opened 2025-04-11
Add unit tests for CLI argument parsing

ginnun:
# Add test cases for configuration handling
Resolves https://github.com/blockfrost/blockfrost-platform/issues/181

This PR enhances the test coverage for the configuration handling in the CLI module by adding several test cases:

- `test_mandatory_ok`: Tests the basic configuration with all required parameters
- `test_mandatory_solitary_ok`: Verifies proper handling of solitary mode configuration
- `test_no_metrics_ok`: Confirms that the no-metrics flag works correctly
- `test_non_defaults_ok`: Tests non-default values for all configuration options
- `test_solitary_conflict_fail`: Verifies that the proper error is raised when conflicting options are provided

Also, now having --solidity and non-solidity arguments give error.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -21,7 +21,7 @@ fn should_skip_serializng_fields<T>(_: &T) -> bool {
     SHOULD_SKIP_SERIALIZNG_FIELDS.load(Ordering::SeqCst)
 }
 
-#[derive(Parser, Debug, Serialize)]
+#[derive(Parser, Debug, Serialize, Clone)]
 #[command(author,
           version = concat!(env!("CARGO_PKG_VERSION"), " (", env!("GIT_REVISION"), ")"),
           about,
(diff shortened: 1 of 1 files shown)


src/cli.rs
```diff
@@ -341,7 +341,15 @@ impl Config {
                 secret,
             })
         } else {
-            None
+            let conflicts = args.reward_address.is_some() || args.secret.is_some();
+
+            if conflicts {
+                return Err(AppError::Server(
+                    "Cannot set --reward-address or --secret in solitary mode (--solitary)".into(),
```
michalrus, line 348: Nice, so you found a _sort of_ a bug!

michalrus (approved):
LGTM – thanks!

vladimirvolek (approved):
nice!

src/cli.rs
ginnun: @michalrus Yeah, let's call it a small improvement ^^

PR #313, opened 2025-05-05
feat(node): Automatically detect the network

vladimirvolek:
- Automatic network detection based on node socket and genesis magic
- Refactored CLI and config
- Added genesis registry + unit tests

Next PR will be custom network setup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 ## Unreleased
 
+### Removed
+
+- `network` parameter from CLI
+
 ### Added
 
 - Load balancing over a WebSocket (eliminating the need for public IP in the future)
-
 - Expose a `health_errors_total` gauge in metrics
 
 ## [0.0.2] - 2025-03-20


ginnun (approved):
lgtm.
Once you merge this I will depend my PR on getting protocol params in this way (instead of going to the node every time)

src/config.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn from_args(args: Args) -> Result<Self, AppError> {
+        Self::from_args_with_detector(args, detect_network)
+    }
+}
+
+fn detect_network(socket_path: &str) -> Result<Network, AppError> {
+    let magics = genesis().all_magics();
+
+    for magic in magics {
+        let ok = Handle::current().block_on(async {
```
michalrus, line 134: From [Slack discussion](https://input-output-rnd.slack.com/archives/C07FD0DNEMQ/p1747904644227909):

```
❯ cargo run -- --server-address 127.0.0.1 --node-socket-path ~/.local/share/blockfrost-platform/preview/node.socket --solitary

[…]

thread 'main' panicked at src/config.rs:134:36:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

src/config.rs
```diff
@@ -0,0 +1,152 @@
...
+        })
+    }
+
+    pub fn from_args(args: Args) -> Result<Self, AppError> {
+        Self::from_args_with_detector(args, detect_network)
+    }
+}
+
+fn detect_network(socket_path: &str) -> Result<Network, AppError> {
+    let magics = genesis().all_magics();
```
michalrus, line 131: Clever! Does `mainnet` appear first? Otherwise we’d get some log noise in `cardano-node`s on Mainnet.

michalrus (commented):
Sorry for the late review, I was out sick:

PR #322, opened 2025-05-26
Implement tx submit evaluation via pallas

ginnun:
The goal of this PR is to implement tx evaluation APIs while trying to stay compatible to the Blockfrost.io version, which are:
- `/utils/txs/evaluate`
- `/utils/txs/evaluate/utxos`

**What is added**
- Two new endpoints and tests
- New cardano node client functions related to ledger: `protocol_params` and `genesis_config`
-`pallas_validate` handling

**What is changed**
- Test folder layout. Now we can have separate files for each endpoint tests. This makes sense since we are going to have lots of them.
- Underlying commit of the pallas tag. I ended up fixing a non-working part in pallas. So changed the hash of the tag. (I want to use tagging better in the future)
- `cargo.lock` file content

Please run `cargo clean` to make sure everything is in a fresh state before compiling.

PR #332, opened 2025-06-30
fix: `aarch64-linux` compilation of Pallas’ `phase2`

michalrus:
## Context

A PR on top of:
* #322 

We had to disable `aarch64-linux` there, because cross-compilation was failing.

Now, that we have `aarch64-linux` machines in CI, let’s try building everything natively there.


michalrus:
@ginnun, okay, it worked locally:

```
$ file result/libexec/blockfrost-platform
result/libexec/blockfrost-platform: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /nix/store/g4agqy9fxhnzmq31idkhp2kqk4sgp3i0-glibc-2.40-66/lib/ld-linux-aarch64.so.1, for GNU/Linux 3.10.0, not stripped
```

I think it should work on CI, too.

So at least the `aarch64-linux` issue will be solved… Windows is left.

michalrus:
The DMG issue is now fixed.

PR #333, opened 2025-06-30
chore: update the devshell for the new Rust and Clippy

michalrus:
## Context

We depend on `ubuntu-latest`, and the latest stable Rust toolchain in the GitHub Actions CI.

It started rejecting our code with the newest release of Clippy, e.g. [/actions/runs/15976292424/job/45059494486](https://github.com/blockfrost/blockfrost-platform/actions/runs/15976292424/job/45059494486).

This PR fixes these rejections, and aligns toolchain versions across CIs.

PR #338, opened 2025-07-01
fix: nix devshell for darwin

vladimirvolek:


michalrus (approved):
Oops, sorry! This is correct.

In a previous PR, [I removed our import of `extra/language/rust.nix`](https://github.com/blockfrost/blockfrost-platform/pull/333/files#diff-a7c83f74c60c9b3c009e923ae941315b7993172f643f48ce8edd365492ac09a4L13), which does this:

https://github.com/numtide/devshell/blob/7c9e793ebe66bcba8292989a68c0419b737a22a0/extra/language/rust.nix#L43-L48

And I forgot to test the devshell on Darwin… 🤦

I removed the import because with our Rust package set there was a conflict between `rustc` provided by `clippy`, and the one provided by `rustc` itself, so I re-added them manually.

PR #344, opened 2025-07-07
Dolos - first part

vladimirvolek:
- Refactored app to crates
- Fetching the binary
- Generating config
- Adjusted CI and docker


Cargo.toml
```diff
@@ -71,11 +69,10 @@ blockfrost = "1.0.5"
...
 [build-dependencies]
-reqwest = { version = "0.12.7", features = ["blocking"] }
-dirs = "5.0"
-bzip2 = "0.4.3"
-tar = "0.4.38"
-zip = "0.6.6"
+build_utils = { path = "crates/build_utils" }
+
+[workspace]
+members = ["crates/common", "crates/dolos", "crates/build_utils", "crates/node"]
```
michalrus, line 75: This split into crates is to prevent cyclic dependencies between modules?

crates/dolos/scripts/download.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
michalrus, line 1: What about Windows?

I know there isn't a binary now, but wouldn’t it be more future-proof to write it in Rust, copying the `build.rs` for `testgen-hs`? :)

crates/dolos/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    // TODO: https://github.com/txpipe/dolos/issues/508
+    if !cfg!(windows) {
+        build_utils::dolos::fetch_binary();
```
michalrus, line 4: So this breaks the Nix build, because its sandbox doesn’t have access to the internet:

```
blockfrost-platform>    Compiling blockfrost-platform v0.0.2 (/build/source)
blockfrost-platform> error: failed to run custom build command for `dolos v0.0.1 (/build/source/crates/dolos)`
blockfrost-platform>
blockfrost-platform> Caused by:
blockfrost-platform>   process didn't exit successfully: `/build/source/target/release/build/dolos-fa84d1c538d6f0e4/build-script-build` (exit status: 101)
blockfrost-platform>   --- stdout
blockfrost-platform>   Downloading dolos version v0.24.0…
blockfrost-platform>
blockfrost-platform>   --- stderr
blockfrost-platform>   bash: /build/source/crates/dolos/scripts/download.sh: No such file or directory
blockfrost-platform>
blockfrost-platform>   thread 'main' panicked at crates/build_utils/src/dolos.rs:88:13:
blockfrost-platform>   build.rs: download script `/build/source/crates/dolos/scripts/download.sh` exited with status exit status: 127
```

I'll fix it by providing Dolos purely, like we do with `TESTGEN_HS_PATH`.

src/main.rs
```diff
@@ -61,6 +55,9 @@ async fn main() -> Result<(), AppError> {
 
     notify_server_ready.notified().await;
 
+    // Start dolos daemon
+    run_daemon(config).await?;
```
michalrus, line 59: I wonder if _we_ should be starting and controlling the Dolos process here?

E.g. we don’t do that for `cardano-node`—we just require that one is running, and to be given a socket path to it.

michalrus (commented):
A few questions, and I'll prepare the Nix fix in the meantime;

PR #352, opened 2025-07-24
Implement unified structure for API endpoints with support for multiple data sources

vladimirvolek:


src/api/pools/extended.rs
```diff
@@ -1,8 +1,7 @@
-use crate::api::ApiResult;
-use axum::Extension;
+use crate::{api::ApiResult, server::state::AppState};
+use axum::extract::State;
 use blockfrost_openapi::models::pool_list_extended_inner::PoolListExtendedInner;
-use dolos::client::Dolos;
 
-pub async fn route(Extension(dolos): Extension<Dolos>) -> ApiResult<Vec<PoolListExtendedInner>> {
-    dolos.pools_extended().await
+pub async fn route(State(state): State<AppState>) -> ApiResult<Vec<PoolListExtendedInner>> {
```
ginnun, line 5: Maybe `PoolListExtendedResponse` could be used? And for the all `route` returns.

ginnun (changes requested):
That's a huge & fantastic refactor 👏🏽

crates/api_provider/src/api.rs
```diff
@@ -0,0 +1,30 @@
...
+/// This struct groups together different implementations of the `ApiProvider` trait,
+/// allowing consumers to call unified methods on a specific backend service via:
+/// `api.dolos.method(...)` or `api.node.method(...)`.
+pub struct Api {
+    pub dolos: Arc<dyn ApiProvider>,
+    // pub node: Arc<dyn ApiProvider>,
+}
+
+impl Api {
+    pub fn new(dolos: Arc<dyn ApiProvider>) -> Self {
```
michalrus, line 27: If all fields are `pub`, then you don't need an `fn new()` (a smart constructor).

crates/api_provider/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    epochs::EpochsApi, genesis::GenesisApi, governance::GovernanceApi, metadata::MetadataApi,
+    network::NetworkApi, pools::PoolsApi, txs::TxsApi,
+};
+use async_trait::async_trait;
+
+/// Unified API interface for all data providers (e.g. Dolos, Node, etc.).
+///
+/// This trait defines a common structure for service-specific implementations.
+/// Each method represents a supported endpoint and can be selectively overridden.
+/// If a method is not implemented, it defaults to returning a 404 via `BlockfrostError::not_found()`.
```
michalrus, line 15: Yea, but we know in compile time what is supported by which API. 🤔

crates/dolos/src/client.rs
```diff
@@ -7,15 +9,18 @@ pub struct Dolos {
...
             let url = Url::parse(&cfg.endpoint).map_err(|e| AppError::Dolos(e.to_string()))?;
-            let client = JsonClient::new(url, cfg.request_timeout)?;
+            let client = JsonClient::new(url, cfg.request_timeout, false)?;
 
-            Ok(Some(Dolos { client }))
+            Ok(Dolos { client })
         } else {
-            // dolos endpoint not configured
-            Ok(None)
+            let is_disabled = true;
```
michalrus, line 19: Hmmm, now I'm thinking, if `JsonClient::is_disabled`, wouldn’t it be cleaner to have an `Option<JsonClient>` instead, without that field? Then you don't have to create fake URLs (that you’ll never use) etc.

ginnun (approved):
lgtm, as far as my comments goes.

PR #361, opened 2025-09-03
Integration test - Icebreakers websockets

vladimirvolek:
resolves: https://github.com/blockfrost/blockfrost-platform/issues/359


tests/endpoints_test.rs
```diff
@@ -282,8 +287,19 @@ mod tests {
...
+
+                let manager = IcebreakersManager::new(
+                    Some(icebreakers_api),
+                    health_errors,
+                    app.clone(),
+                    api_prefix,
+                );
+
+                tokio::spawn(async move {
+                    manager.run().await;
```
michalrus, line 301: Nice, but will the test fail if there's something wrong with the WebSocket? Looking at `IcebreakersManager::run`, it will just continue to retry.

Hmmm, and this test will end, right after `tokio::spawn`, killing the WebSocket task – we don’t join it in any way.

michalrus: Maybe the new manager should have two methods:
* `run` — for production, and it does retry,
* `run_once` — that is repeatedly called `run`, and in the tests?

Then you can more easily wait for (join) `run_once` and check that it succeeded.

michalrus (commented):
Very nice! I wrote a lot of comments, but they boil down to the fact that `icebreakers_api` doesn’t need to be an `Option<_>`, looking at how it's used — it’s _always_ `Some(_)` in the manager.

PR #362, opened 2025-09-03
chore: run `blockfrost-tests` on CI against the Platform

### LINKED ISSUE #327: New CI workflow to run `blockfrost-tests` against the Platform
## Context
 
* See this [Slack thread](https://input-output-rnd.slack.com/archives/C07FD0DNEMQ/p1748872238900269).
* See another [Slack thread about `nix3-run`](https://input-output-rnd.slack.com/archives/C07FD0DNEMQ/p1748537954753149).

* The tests: [blockfrost/blockfrost-tests](https://github.com/blockfrost/blockfrost-tests).

## Plan

We have to run these tests on our own self-hosted `github-runner`, to access the Preview `cardano-node` and `dolos`.

In this case, we can use a `nix3-run` script that calls `blockfrost-platform` built by Nix. Most of the time, the Cargo dependencies would be cached in `/nix/store` of that runner machine, making the build fast.

It would also allow easy local running.

A downside (or upside, depending how you look) is that we'd pin a concrete Git revision of `blockfrost-tests`. But if this is eventually considered to be a downside, we could also `git-clone` them each time in the `nix3-run` script.

michalrus:
Resolves #327

## Context

* #327

* I had to change some `Extension(config)` that were missed to `State(app_state)` – this made more tests pass.

* Currently we run them in parallel with `integration_tests` on that same Preview runner.

* I had to add `name = "blockfrost-platform"` to the CLI config, otherwise `--version` reported as `common 1.2.3 deadbeef` instead of `blockfrost-platform 1.2.3 deadbeef`.

* All necessary secrets have been added to this repo’s config.

## Results

On Dolos 0.26.0 on CI, [we're getting 29.5%](https://github.com/blockfrost/blockfrost-platform/actions/runs/17447239144/job/49544840625?pr=362) in about 1 minute:

```
      Tests  203 failed | 85 passed (288)
```


vladimirvolek (approved):
Very very good, thanks!

PR #376, opened 2025-09-09
update nextra

vladimirvolek:


michalrus (approved):
I'm going to approve, but I'm really not sure about committing binary blobs ([Yarn Zero-installs](https://yarnpkg.com/features/caching#zero-installs)) to the repository, especially since we’re more/less backing up the Internet. 😃

E.g. this update has grown it by 6 MB, and we’re stuck with it till the end of this project:

```
❯ git rev-list --objects 536ebcf --not '536ebcf^@' \
| awk '{print $1}' \
| git cat-file --batch-check='%(objecttype) %(objectsize)' \
| awk '$1=="blob"{s+=$2} END{printf "%.2f MB\n", s/1024/1024}'

5.97 MB
```

Quoting [github/git-sizer](https://github.com/github/git-sizer):

> Ideally, Git repositories should be under 1 GiB, and (without special handling) they start to get unwieldy over 5 GiB. Big repositories take a long time to clone and repack, and take a lot of disk space.

1 GB is ~170 such updates, not very few, but not very many either.

@ginnun maybe you have some thoughts?

PR #378, opened 2025-09-12
Tx evaluation

### LINKED ISSUE #306: Add `POST /utils/txs/evaluate` – how many execution units for a transaction
## Context

* One of Blockfrost’s most frequently used endpoints:
  * https://docs.blockfrost.io/#tag/cardano--utilities/POST/utils/txs/evaluate
  * https://docs.blockfrost.io/#tag/cardano--utilities/POST/utils/txs/evaluate/utxos

* Reference implementation is [available in Ogmios](https://github.com/CardanoSolutions/ogmios/blob/master/architectural-decisions/accepted/015-script-execution-units-evaluator.md).

* It’s notoriously difficult to get right – see [this `#ledger` thread on Slack](https://input-output-rnd.slack.com/archives/CCRB7BU8Y/p1744273705279819).
 
* Our [Slack discussion](https://input-output-rnd.slack.com/archives/C07FD0DNEMQ/p1744377128700929).

## How?

We could use a similar approach as with the CBOR error deserializer. I.e.:

1. Generate thousands of random but more/less sound transactions.
1. Compare the calculation of Ogmios / Cardano CLI with our implementation for each transaction.

Both the transactions and the "official" calculation would similarly come from [`testgen-hs` in Haskell](https://github.com/input-output-hk/testgen-hs).

The generator part is tracked in:
* #320

ginnun:
Resolves #306

**Stacked on #537** - merge that one first.

## Context

Implements tx evaluation APIs compatible with the Blockfrost.io version:

- `/utils/txs/evaluate`
- `/utils/txs/evaluate/utxos`

## What is added

- Two new endpoints with integration tests
- `bf-tx-evaluator` crate providing native (Rust, via `pallas-validate`) and external (Haskell, via `testgen-hs`) evaluators
- `ChainConfigWatch` in `bf-node` - a background task that watches the Cardano node and publishes protocol params + genesis config via `tokio::sync::watch`, refreshed only at epoch boundaries
- Node helpers: `ledger`, `utxo`, `chain_config_watch`

## What is changed

- Pallas fork switched to `blockfrost/pallas` with `pallas-validate` (`phase2` feature) for script evaluation
- CBOR validation error messages improved (graceful fallback on `pallas-hardano` formatting failure)
- Devshell picks up tools needed by the new native deps - m4 (for `secp256k1-sys`), make/awk/diffutils (for integration tests)
- `cardano-node` trace silencing migrated to the newer API
- CI: `CARGO_PROFILE_DEV_DEBUG=0` on integration-test jobs to reduce disk pressure

## What is scoped out (now in #537)

- `testgen-hs` subprocess handling extracted from `crates/node/src/cbor/` into dedicated `bf-testgen` and `bf-error-decoder` crates
- `testgen-hs` version bump to `10.6.3.1` and release binary downloads switched to `blockfrost/testgen-hs`
- `build_utils` target detection helpers factored out


nix/internal/unix.nix
```diff
@@ -163,14 +166,15 @@ in
     in
       (import inputs.flake-compat {
         src =
+          if targetSystem != "aarch64-darwin" && targetSystem != "aarch64-linux"
```
michalrus, line 169: This looks like a merge/rebase artifact – wrong conflict resolution? The line is repeated.

nix/internal/unix.nix
```diff
@@ -220,7 +224,24 @@ in
...
+      (import inputs.flake-compat {
+        src =
+          if targetSystem != "aarch64-linux"
+          then unpatched
+          else {
+            outPath = toString (pkgs.runCommand "source" {} ''
+              cp -r ${unpatched} $out
+              chmod -R +w $out
+              cd $out
+              patch -p1 -i ${./testgen-hs--enable-aarch64-linux.diff}
```
michalrus, line 239: Wait, but we already have `aarch64-linux` on `main`. Something is off here, another conflict?

src/api/utils/txs/evaluate/model.rs
```diff
@@ -0,0 +1,359 @@
...
+                    "script": {"plutus:v2": "base16script"}
+                  }
+                ]
+                ]
+        });
+
+        let _req: TxEvaluationRequest = serde_json::from_value(value).unwrap();
+    }
+}
+*/
```
michalrus, line 359: Should this file be commented out and committed?

It also looks like we don't have `treefmt` run on this, since it's missing a trailing newline.

crates/testgen/src/testgen.rs
```diff
@@ -11,19 +11,22 @@ use tokio::sync::{mpsc, oneshot};
 use tracing::{debug, error, info};
 
 #[derive(Clone)]
-pub struct FallbackDecoder {
-    sender: mpsc::Sender<FDRequest>,
-    current_child_pid: Arc<AtomicU32>,
+pub struct Testgen {
+    pub sender: mpsc::Sender<TestgenRequest>,
```
michalrus, line 15: Why do these internal fields need to be public for everyone?

crates/testgen/src/testgen.rs
```diff
@@ -143,36 +151,6 @@ impl FallbackDecoder {
         ))
     }
 
-    /// This function is called at startup, so that we make sure that the worker is reasonable.
-    pub async fn startup_sanity_test(&self) -> Result<(), String> {
```
michalrus, line 147: But isn’t this useful? Just in case that someone plugs another binary into the `libexec` directory.

michalrus (commented):
Hey, I started reviewing, but then noticed that it's a draft! Left a few comments either way :)

crates/testgen/src/testgen.rs
ginnun: This uses error serialization/deserialization. I can expect every binary integration to have a startup_sanity_test implementation of their own.
For evaluation, it's already handled via the init functionilty.

src/api/utils/txs/evaluate/model.rs
ginnun: Fixed in b4b535e552d52a2541d81728db47811dc9b1ace8

crates/testgen/src/testgen.rs
ginnun: Changed in bbec0dbf45875089fe67ec1e08faf79ab47aacac

crates/testgen/src/testgen.rs
michalrus: I was thinking more like… What if the user puts a different binary there? It was more of a sanity test from the point of view of the `blockfrost-platform`, i.e. "is this a reasonable helper binary that we expect" to rule out obvious user errors

michalrus (commented):
I pushed another merge with `main`, because there were some more conflicts to resolve. I hope I did that correctly!

Now I see that `cargo test` tests are failing:

```
❯ cargo nextest run --workspace --lib --verbose

…

     Summary [  76.650s] 650 tests run: 195 passed, 455 failed, 0 skipped
```

Stuff like:

```
        FAIL [   1.058s] blockfrost-platform-error-decoder external::tests::test_cbor_0001_ConwayTreasuryValueMismatch_Coin_Coin
```

Or:

```
        FAIL [   1.086s] blockfrost-platform-error-decoder tests::specific::test_cbor_0013_ConwayWdrlNotDelegatedToDRep_KeyHash_KeyHash
```

michalrus:
> The plan is to fix/update the code for `pallas/hardano`.

Great! Here's a tracking issue for this new work:
* #459

nix/internal/unix.nix
```diff
@@ -191,9 +191,28 @@ in
...
+    in {
+      cardano-node = mkTool "cardano-node";
+      cardano-cli = mkTool "cardano-cli";
+      cardano-submit-api = mkTool "cardano-submit-api";
+    };
+
     cardano-node-packages =
       {
-        x86_64-linux = cardano-node-flake.hydraJobs.x86_64-linux.musl;
+        x86_64-linux = cardano-node-release-linux;
```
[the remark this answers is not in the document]
michalrus: And since the tests are passing – this would mean that SREs updated their node on our GitHub Actions runner.

nix/internal/unix.nix
```diff
@@ -188,9 +191,28 @@ in
...
```
michalrus: I'm fairly certain that this is not necessary in this PR, let's verify again.

See these comments: https://github.com/blockfrost/blockfrost-platform/pull/378#discussion_r2872247931.

michalrus (changes requested):

nix/internal/unix.nix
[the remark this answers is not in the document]
michalrus: I think I get it now – please see https://github.com/blockfrost/blockfrost-platform/pull/378#discussion_r2872404626

michalrus (commented):
WDYT of this last change before a proper review, so that `testgen-hs` version is in sync with `cardano-node` in the shell? Hopefully it's only a formality, but worth checking, IMO.

michalrus (changes requested):

.github/workflows/ci.yaml
```diff
@@ -127,6 +127,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Free disk space
+        run: nix-collect-garbage -d
```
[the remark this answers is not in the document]
hlolli: It's correct that the runner doesn't always clean the repos after pulling them. Deleting build folder could help a little if it's big. I think the checkout action pulls just 1 commit and doesn't store history, but doesn't remove the repo afterwards, but removes and cleans on fetch

image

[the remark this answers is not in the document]
ginnun: > Other PRs don’t fill it, so maybe it's testgen-hs somehow?
@michalrus   It's the Rust debug build artifacts accumulating in the runner's /run/github-runner/.../target/debug/deps/ directory (after my /nix/store clean), not testgen-hs.
                                                                                                                                                                                                                                              
  This PR adds pallas_validate as a new transitive dependency (via tx_evaluator). In debug mode (default debug = 2), its .rlib is hundreds of MB. From what I understand (i might be wrong), the self-hosted runner never cleans target/, so old hashed artifacts from previous runs pile up.
  The linker failure tells that it dies (just after tx_evaluator/pallas_validate) trying to write the final test binary after the Nix GC already freed 16 GB:
  ld.bfd: final link failed: No space left on device

  Temporary fix to check the theory: clean target/debug/ before building + set CARGO_PROFILE_DEV_DEBUG=0 (maybe later) to keep future artifacts small. Working on it.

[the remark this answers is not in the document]
michalrus: @hlolli, @cleverca22, I mean, but should we then be running the builds inside `/run/` (on `tmpfs`, in limited RAM), instead of somewhere on the (more unlimited) SSD?

Maybe the GitHub runner service is configured in a strict systemd sandbox which defaults to `/run`? Or are its directories in `/run` just symlinks to somewhere on the SSD?

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,893 @@
...
+        Ok(report) => Ok(report),
+        Err(e) => Err(BlockfrostError::custom_400(
+            "Error evaluating transaction: ".to_string() + &e.to_string(),
+        )),
+    }
+}
+
+pub fn create_address(addr: &str) -> Bytes {
+    Address::from_bech32(addr)
+        .unwrap_or_else(|_| Address::from_hex(addr).unwrap())
```
michalrus: User-supplied strings should not cause `panic!`s

crates/platform/src/api/utils/txs/evaluate/utxos.rs
```diff
@@ -1,6 +1,102 @@
...
+                            tx_cbor.as_slice(),
+                            request.additional_utxo_set,
+                        )
+                        .await?
+                } else {
+                    let r = convert_eval_report_v5(
+                        evaluate_binary_tx(node, tx_cbor.as_slice(), request.additional_utxo_set)
+                            .await?,
+                    );
+                    serde_json::to_value(r).unwrap()
```
michalrus: Can this `panic!`? Let's use `?`, too

crates/platform/src/server/routes/hidden.rs
```diff
@@ -135,9 +136,9 @@ pub fn get_hidden_api_routes(enable_metrics: bool) -> Router<AppState> {
         .route("/txs/{hash}/required_signers", get(txs::hash::required_signers::route))
         .route("/txs/{hash}/cbor", get(txs::hash::cbor::route))
 
-        // utils
-        .route("/utils/tx/evaluate", post(utils::txs::evaluate::root::route))
-        .route("/utils/tx/evaluate/utxos", post(utils::txs::evaluate::utxos::route));
+        // tx evaluate
+        .route("/utils/tx/evaluate",post(utils::txs::evaluate::root::route))
```
michalrus: The original legacy Blockfrost API uses plural here (`/txs/`, not `/tx/`):

https://docs.blockfrost.io/#tag/cardano--utilities/POST/utils/txs/evaluate

crates/tx_evaluator/src/model/api.rs
```diff
@@ -0,0 +1,1000 @@
...
+    StartsAt(u64),
+    #[serde(untagged)]
+    NOf(HashMap<String, Vec<ScriptNative>>),
+    #[serde(untagged)]
+    Signature(String),
+}
+
+// JSON response
+#[derive(Serialize)]
+pub struct TxEvaluationResponse {
```
michalrus: Is it enough that it only contains `spend:1` and `mint:0`?

Also: dead code?

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,893 @@
...
+
+//* This implementation uses pallas validate.
+//  Since pallas validate behaves differently from the ogmios validation (which uses ledger)
+//  this implementation is not used at the moment  */
+pub async fn evaluate_binary_tx(
+    node_pool: NodePool,
+    tx_cbor_binary: &[u8],
+    utxos: Option<AdditionalUtxoSet>,
+) -> Result<EvalReport, BlockfrostError> {
+    let slot_config = pallas_validate::phase2::script_context::SlotConfig::default();
```
michalrus: Could it be that the default `SlotConfig` wouldn’t match the network’s?

michalrus (changes requested):
Ahh, it’s a large one! 🐳

Some comments below.

Can we also add the two new endpoints to the ones we run `blockfrost-tests` against in `supported_endpoints.json`? It will be automatically picked up by the CI, and we can see the change:

https://github.com/blockfrost/blockfrost-platform/blob/db42ec026433a6958e608c2a4bd2130880417b3a/crates/platform/tests/data/supported_endpoints.json#L1-L4

crates/common/src/chain_config.rs
```diff
@@ -0,0 +1,76 @@
...
+            zero_slot: 0,
+            zero_time: 1666656000000,
+            epoch_length: 86400,
+        }
+    }
+
+    pub fn by_network_magic(network_magic: &u32) -> Self {
+        match network_magic {
+            764824073 => Self::preview(),
+            1 => Self::mainnet(),
```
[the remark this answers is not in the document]
ginnun: Yeah, how did I do this? These are only used in slot calculations, maybe this is why they slipped. 

I couldn't find a common place to use them from.

Fixed in bdad9bd356669dc66c98fc679906fcd8f5991f39 .

crates/common/src/errors.rs
```diff
@@ -70,6 +70,12 @@ impl From<hex::FromHexError> for BlockfrostError {
     }
 }
 
+impl From<serde_json::Error> for BlockfrostError {
+    fn from(err: serde_json::Error) -> Self {
+        Self::internal_server_error(format!("SerdeError: {err}"))
```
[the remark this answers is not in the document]
ginnun: Fixed in 3e17ec8.

crates/common/src/chain_config.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl ChainConfigCache {
+    pub fn new(genesis_config: GenesisConfig, protocol_params: CurrentProtocolParam) -> Self {
+        let slot_config = SlotConfig::by_network_magic(&genesis_config.network_magic);
+
+        Self {
+            genesis_config,
+            protocol_params,
+            slot_config,
+            era: 6, // conway
```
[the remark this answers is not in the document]
ginnun: Added a comment in 7cebed7.

crates/tx_evaluator/src/ledger.rs
```diff
@@ -0,0 +1,11 @@
+pub enum TransactionScriptFailure {
```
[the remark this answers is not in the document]
ginnun: Removed in 3e1748f.

.gitignore
```diff
@@ -31,6 +31,7 @@ result-*
 
 /cardano-node-configs
 /cardano-node
+/data
```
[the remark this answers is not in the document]
ginnun: I used this folder for cardano and dolos node data (the way I prefer), also as an output folder for various bash commands.
Once creates a problem, this kind of things can always be changed on spot.
Removed it in ab57e66 anyways.

crates/testgen/src/testgen.rs
```diff
@@ -85,52 +124,92 @@ impl FallbackDecoder {
...
+            search_paths.push(
+                target_dir
+                    .join("testgen-hs")
+                    .join("extracted")
+                    .join(exe_name),
+            );
+        }
+
+        // Compile-time CARGO_MANIFEST_DIR always points to this crate.
+        let target_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join(target_dir_from_env);
```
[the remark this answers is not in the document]
ginnun: Applied in 56996fc.

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,893 @@
...
+            .iter()
+            .map(convert_byron_txin)
+            .collect(),
+        MultiEraTx::Conway(x) => x
+            .transaction_body
+            .inputs
+            .iter()
+            .map(convert_alonzo_txin)
+            .collect(),
+        _ => unreachable!("unknown era transaction"),
```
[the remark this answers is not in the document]
ginnun: Converted into a 400 error in e43824c4b898f709bf3a1949102133a21cf718d1.

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,915 @@
...
+                AnyUInt::U64(ada.lovelace),
+                NonEmptyKeyValuePairs::from_vec(assets_mut).unwrap(),
+            )
+        },
+    }
+}
+
+fn parse_asset_string(str: &str) -> (PolicyId, AssetName) {
+    let mut parts = str.split('.');
+    let policy_id = parts.next().expect("Invalid asset string in utxo output");
```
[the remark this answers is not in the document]
ginnun: @hlolli Are you suggesting to keep the line as is?

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,893 @@
...
+        >(&tx_out_cbor)
+        .unwrap();
+        let tx_out = pallas_primitives::conway::TransactionOutput::<'_>::PostAlonzo(post_alonzo);
+        let multi_era_out = MultiEraOutput::<'_>::Conway(Box::new(Cow::Owned(tx_out)));
+
+        utxos.insert(TxoRef::from(&multi_era_in), EraCbor::from(multi_era_out));
+    }
+    let mut node = node_pool.get().await?;
+
+    let txins = extract_inputs(multi_era_tx.clone());
```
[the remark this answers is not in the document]
ginnun: Fixed in cb3cad6.

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,893 @@
...
+            tx_out.script.map(|script| CborWrap(script.into()));
+
+        let post_alonzo = pallas_primitives::conway::PostAlonzoTransactionOutput {
+            address,
+            value,
+            datum_option,
+            script_ref,
+        };
+
+        let tx_out_cbor = pallas_codec::minicbor::to_vec(post_alonzo).unwrap();
```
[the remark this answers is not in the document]
ginnun: These `unwraps` won't fail since they are round-trips and already valid pallas data.
To create a `KeepRaw<>` object, you need to provide raw bytes, so we are round-tripping.
But to get rid of future suspicions, applied your feedback in 8738486ff3bb5ac26b963b6ac94dd46bb3f3d9f1.

crates/tx_evaluator/src/native.rs
ginnun: Derived SlotConfig from genesis in e1be8f1.

crates/platform/src/api/utils/txs/evaluate/utxos.rs
```diff
@@ -1,6 +1,102 @@
...
+    Extension(node): Extension<NodePool>,
+    Extension(fallback_evaluator_opt): Extension<Option<ExternalEvaluator>>,
+    headers: HeaderMap,
+    Query(query): Query<EvaluateQuery>,
+    extract::Json(tx_request): extract::Json<TxEvaluationRequest>,
+) -> Result<impl IntoResponse, BlockfrostError> {
+    // Allow only application/json content type
+    validate_content_type(&headers, &["application/json"])?;
+
+    let version: u8 = query.version.parse().unwrap();
```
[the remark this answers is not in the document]
ginnun: Handled in 8738486ff3bb5ac26b963b6ac94dd46bb3f3d9f1.

crates/tx_evaluator/src/model/api.rs
```diff
@@ -0,0 +1,1000 @@
...
+            ScriptV6::PlutusV1 { cbor } => ScriptRef::PlutusV1Script(PlutusScript::<1>(
+                Bytes::from(hex::decode(cbor).unwrap()),
+            )),
+            ScriptV6::PlutusV2 { cbor } => ScriptRef::PlutusV2Script(PlutusScript::<2>(
+                Bytes::from(hex::decode(cbor).unwrap()),
+            )),
+            ScriptV6::PlutusV3 { cbor } => ScriptRef::PlutusV3Script(PlutusScript::<3>(
+                Bytes::from(hex::decode(cbor).unwrap()),
+            )),
+            ScriptV6::Native { cbor, json } => cbor.clone().map_or_else(
```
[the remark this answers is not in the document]
ginnun: cbor and json coming from pallas is not possible. Although there are both `Option`, it's in xor fashion.

But should you trust an external lib?

Handled the worst case in d6f0b64455974f75d1a4e5b799b739cbadad22ac.

crates/tx_evaluator/src/wrapper.rs
```diff
@@ -0,0 +1,145 @@
...
+    response: Vec<TxEvalResultV5>,
+    mirror: serde_json::Value,
+) -> serde_json::Value {
+    // flatten objects
+    let mut result_map = Map::with_capacity(response.len());
+    for r in response {
+        serde_json::to_value(r)
+            .unwrap()
+            .as_object()
+            .unwrap()
```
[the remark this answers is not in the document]
ginnun: I am not sure how can these already type-safe & serde decorated data can fail? 
But still, I converted them to more descriptive `.expect` 00fc720

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,262 @@
...
+                "ExternalEvaluator: failed to serialize initial payload: {err}"
+            ))
+        })?;
+
+        /*
+        (self.testgen.send(payload).await).map_err(|err| {
+            AppError::Server(format!("ExternalEvaluator: Failed to initialize: {err}"))
+        })
+        */
+        // println!("1 -> {}", payload);
```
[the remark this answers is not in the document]
ginnun: Removed in 00fc720.

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,262 @@
...
+        let testgen = Testgen::spawn("evaluate-stream")
+            .map_err(|err| AppError::Server(format!("Failed to spawn ExternalEvaluator: {err}")))?;
+
+        let evaluator = Self { testgen };
+        evaluator.init(config).await?;
+
+        Ok(evaluator)
+    }
+
+    /// Sends repeatative data as the first communication so we don't need to send every time.
```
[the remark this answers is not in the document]
ginnun: Corrected in 00fc720.

crates/tx_evaluator/src/model/api.rs
```diff
@@ -0,0 +1,1000 @@
...
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct TxEvalFailureV6 {
+    pub validator: TxValidator,
+    pub error: serde_json::Value,
+}
+
+#[derive(Serialize, Deserialize)]
+pub enum TxValidatorV6 {
```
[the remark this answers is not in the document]
ginnun: Yes, we will get only these two in an evaluation response.
I removed this.

crates/tx_evaluator/src/model/api.rs
```diff
@@ -0,0 +1,1000 @@
...
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(untagged)]
+pub enum EvaluationError {
+    Evaluation(serde_json::Value),
+    Deserialization(DeserializationErrorData),
+}
+
+pub enum EvaluateTransactionError {}
```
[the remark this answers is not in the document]
ginnun: Removed in 44e7945.

crates/error_decoder/src/external.rs
```diff
@@ -0,0 +1,1537 @@
...
+            Ok(())
+        } else {
+            Err(format!(
+                "ExternalDecoder: startup_sanity_test failed: {result:?}"
+            ))
+        }
+    }
+
+    #[cfg(test)]
+    /// A single global [`ExternalDecoder`] that you can cheaply use in tests.
```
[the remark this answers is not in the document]
ginnun: Handled in bdad9bd.

crates/tx_evaluator/src/wrapper.rs
```diff
@@ -0,0 +1,145 @@
...
+        "result": result_map,
+        "reflection": mirror,
+    })
+}
+
+pub fn wrap_response_v6(resp: TestgenResponse, id: serde_json::Value) -> serde_json::Value {
+    match resp {
+        TestgenResponse::Ok(value) => {
+            println!("value6 {:?}", value);
+            let decoded: Vec<TxEvalResultV6> = serde_json::from_value(value).unwrap();
```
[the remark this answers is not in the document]
ginnun: TestgenResponse::ok already contains a `serde::value` here, so there should be no way for it to fail serialization.
I used more descriptive `.expect` in https://github.com/blockfrost/blockfrost-platform/commit/00fc720ba3ac5e381c2732b339f31cf35b875332

crates/tx_evaluator/src/wrapper.rs
```diff
@@ -0,0 +1,145 @@
...
+
+            if is_success_v6(&decoded) {
+                wrap_success_response_v6(decoded, id)
+            } else {
+                {
+                    json!({
+                        "jsonrpc": "2.0",
+                        "method": "evaluateTransaction",
+                        "error": "hello error",
```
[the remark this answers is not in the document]
ginnun: Fixed in [c901a81](https://github.com/blockfrost/blockfrost-platform/pull/378/commits/c901a81).

crates/build_utils/src/target.rs
```diff
@@ -0,0 +1,21 @@
+pub fn os() -> &'static str {
```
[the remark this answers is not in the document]
ginnun: Looks like a rebase/merge artifact. Deleted in [f269525](https://github.com/blockfrost/blockfrost-platform/pull/378/commits/f269525).

crates/tx_evaluator/src/model/api.rs
```diff
@@ -0,0 +1,1000 @@
...
+    pub memory: u64,
+    pub cpu: u64,
+}
+
+// Create a wrapper type for TxEvalResult
+
+#[derive(Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum TxEvalResultV5 {
+    SUCCESS(TxEvalSuccessV5),
```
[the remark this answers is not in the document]
ginnun: Corrected in 00fc720ba3ac5e381c2732b339f31cf35b875332.

crates/platform/tests/common/mod.rs
```diff
@@ -30,6 +31,14 @@ pub fn initialize_logging() {
     let _ = INIT_LOGGING;
 }
 
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
ginnun: Removed in https://github.com/blockfrost/blockfrost-platform/commit/e15aa7dc6dc2be4ba4c89d5016d3d2fbd7b66415.

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,915 @@
...
+            assets: None,
+        } => pallas_primitives::conway::Value::Coin(*coins),
+        Value {
+            coins,
+            assets: Some(assets_map),
+        } => {
+            let mut assets = BTreeMap::new();
+            for (id_name, number) in assets_map {
+                let mut asset_detail = BTreeMap::new();
+                let coin: PositiveCoin = number
```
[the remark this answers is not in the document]
hlolli: No, endpoint is reachable via query override; malformed assets/policy IDs can panic.
I could suggest we replace expect/unwrap in this path with fallible validation (map_err(...)?) and return BlockfrostError::custom_400(...) for malformed additionalUtxo fields (asset id/name/amount).
If this is experimental, you can ignore this comment :)

crates/tx_evaluator/src/wrapper.rs
```diff
@@ -0,0 +1,155 @@
...
+ * - failed eval: top-level 'result' field with error json details
+ * - faulty input: top-level 'fault' field with error string
+ */
+pub fn wrap_response_v5(resp: TestgenResponse, mirror: serde_json::Value) -> serde_json::Value {
+    //println!("wrap_response_v5: {:?}", resp);
+    match resp {
+        TestgenResponse::Ok(value) => {
+            // The external evaluator returns a v6-like structure, but we need to convert it to v5.
+            let v6_results: Vec<TxEvalResultV6> =
+                serde_json::from_value(value).expect("ExternalEvaluator: invalid v6 result format");
```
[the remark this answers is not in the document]
hlolli: no a big deal, more descriptive .expect is also good for debugging and reliability

crates/tx_evaluator/src/model/api.rs
```diff
@@ -0,0 +1,1005 @@
...
+            ScriptNative::Any(scripts) => {
+                NativeScript::ScriptAny(scripts.into_iter().map(|s| s.into()).collect())
+            },
+            ScriptNative::All(scripts) => {
+                NativeScript::ScriptAll(scripts.into_iter().map(|s| s.into()).collect())
+            },
+            ScriptNative::ExpiresAt(time) => NativeScript::InvalidHereafter(time),
+            ScriptNative::StartsAt(time) => NativeScript::InvalidBefore(time),
+            ScriptNative::NOf(h_map) => {
+                let (n_str, scripts) = h_map.into_iter().next().expect("NOf: empty map");
```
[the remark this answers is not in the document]
hlolli: To elaborate: my concern is not about valid requests, but about what happens on malformed request bodies in the runtime request path.

In this code, `ScriptNative::NOf` is modeled as `HashMap<String, Vec<ScriptNative>>`, and then later converted with:

 ```rust
 let (n_str, scripts) = h_map.into_iter().next().expect("NOf: empty map");
 n_str.parse::<u32>().expect("NOf: invalid n key")
 ```

 That means some bad payloads can make us panic instead of returning a deterministic `400`.

 Concrete examples:

 ```json
 { "native": {} }
 ```
 can hit `expect("NOf: empty map")`

 ```json
 { "native": { "foo": [] } }
 ```
 can hit `expect("NOf: invalid n key")`

 and similarly nearby helpers like `decode_pubkey_hash` / `decode_script_hex` use `expect(...)`, so malformed key hashes or script CBOR can also panic.

 Since this is request parsing/normalization for client input, I’d prefer either:

 1. reject these forms during deserialization, or
 2. make the conversion fallible (`TryFrom` / `Result<..., BlockfrostError>`) and map invalid input to `custom_400(...)`.

The main point is: invalid client payload should become a controlled `400`, not a process panic / `500`.

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,267 @@
...
+                                    "invalid tx_id '{}': {e}",
+                                    utxo.tx_id
+                                ))
+                            })?,
+                        index: AnyUInt::U64(utxo.index),
+                    };
+                    Ok(Self::build_transaction_output(
+                        txin,
+                        create_address(&tout.address)?,
+                        tout.script
```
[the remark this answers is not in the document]
hlolli: perfect

hlolli (approved):

crates/common/src/chain_config.rs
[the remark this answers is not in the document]
michalrus: Out of curiosity, what happens if:
- the node is still syncing, and say, still in Babbage?
- they add era № 7, will we fail silently?

.gitignore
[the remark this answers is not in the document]
michalrus: For a personal Git ignore, you also have:
- `~/.config/git/ignore` – global,
- `.git/info/exclude` – in this repo

crates/node/src/cbor/tests/test_data.rs
```diff
@@ -0,0 +1,185 @@
+use pallas_primitives::{
```
michalrus: This file is never included in compilation?

crates/platform/src/api/utils/txs/evaluate/utxos.rs
```diff
@@ -1,6 +1,75 @@
...
-    Err(BlockfrostError::not_found())
+pub async fn route(
+    Extension(node): Extension<NodePool>,
+    Extension(fallback_evaluator): Extension<Option<ExternalEvaluator>>,
+    headers: HeaderMap,
+    Query(query): Query<EvaluateQuery>,
+    extract::Json(tx_request): extract::Json<TxEvaluationRequest>,
+) -> Result<impl IntoResponse, BlockfrostError> {
+    // Allow only application/json content type
+    validate_content_type(&headers, &["application/json"])?;
```
michalrus: Is this necessary? We already define the parameter as `extract::Json(tx_request): extract::Json<TxEvaluationRequest>`, so we won’t even get to this line unless it was indeed JSON.

michalrus (changes requested):
A couple more comments:

crates/testgen/src/testgen.rs
```diff
@@ -0,0 +1,325 @@
...
+                let single_run = Self::spawn_child(
+                    &testgen_hs_path_for_thread,
+                    &mut receiver,
+                    &mut last_unfulfilled_request,
+                    &current_child_pid_clone,
+                    &variant_clone,
+                );
+                let restart_delay = std::time::Duration::from_secs(1);
+                error!(
+                    "Testgen: will restart in {restart_delay:?} because of a subprocess error: {}",
```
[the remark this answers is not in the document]
ginnun: When it comes to tx evaluation, It's not a fallback anymore. It's an external evaluator, The name "external" is better in my opinion. Whatever we use it as a fallback or a main is up to us and should not be enforced in the name Hence, it's now the binaries name.

michalrus (changes requested):
One quick comment about the other repo (Pallas), and now I'll process the rest (here):

crates/tx_evaluator/src/model/api.rs
```diff
@@ -0,0 +1,1041 @@
...
+            allegra: err.clone(),
+            mary: err.clone(),
+            alonzo: err.clone(),
+            babbage: err.clone(),
+            conway: err,
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for DeserializationErrorData {
```
michalrus: Hmmmm, you're _deriving_ `Serialize` above:

https://github.com/blockfrost/blockfrost-platform/blob/fe4dd66ee371471e784202e9b94cfb1c15627000/crates/tx_evaluator/src/model/api.rs#L184-L185

… so after this generic serialization, it ends up as JSON object with 6 fields.

But here, it seems like you're deserializing a single string?

So serializer is not a inversion of deserializer, and they don’t match? I think in general JSON codecs should be invertible, or it’s surprising.

michalrus (changes requested):
Okay – another round done! The most important comment is about `blockfrost-tests`, I think.

crates/platform/tests/data/supported_endpoints.json
```diff
@@ -37,5 +37,7 @@
   "/epochs/latest/parameters",
   "/genesis",
   "/addresses/{address}/transactions",
-  "/addresses/{address}/utxos"
+  "/addresses/{address}/utxos",
+  "/utils/txs/evaluate",
```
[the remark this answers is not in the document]
michalrus: And you can easily run a decent Dolos with `run-dolos-preview` inside the devshell.

But since you don’t need to test any Dolos integration, you can just as well comment it out!

Cargo.toml
```diff
@@ -52,13 +57,16 @@ metrics-exporter-prometheus = { version = "0.18.1", default-features = false }
...
-pallas-codec = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha4" }
-pallas-crypto = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha4" }
-pallas-hardano = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha4" }
-pallas-network = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha4" }
-pallas-primitives = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha4" }
-pallas-traverse = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha4" }
+pallas-addresses = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha6" }
+pallas-codec = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha6" }
+pallas-crypto = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha6" }
+pallas-hardano = { git = "https://github.com/blockfrost/pallas.git", tag = "blockfrost-platform-0.0.3-alpha6" }
```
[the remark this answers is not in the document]
ginnun: @michalrus https://github.com/txpipe/pallas/pull/624 is ready.

ginnun (commented):
@michalrus I replied, but kept them in `pending` for a while ^^

crates/tx_evaluator/src/wrapper.rs
```diff
@@ -0,0 +1,165 @@
...
+    })
+}
+
+pub fn wrap_response_v6(resp: TestgenResponse) -> serde_json::Value {
+    let id = serde_json::json!(uuid::Uuid::new_v4().to_string());
+    match resp {
+        TestgenResponse::Ok(value) => {
+            let decoded: Vec<TxEvalResultV6> =
+                serde_json::from_value(value).expect("ExternalEvaluator: invalid v6 result format");
+
```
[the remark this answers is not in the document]
ginnun: Fixed in 359d8b12.

crates/platform/src/server.rs
```diff
@@ -61,6 +62,12 @@ pub async fn build(
     // Set up optional Icebreakers API (solitary option in CLI)
     let icebreakers_api = IcebreakersAPI::new(&config, api_prefix.clone()).await?;
 
+    // Initialize chain configurations
+    let chain_config_cache = init_caches(node_conn_pool.clone()).await?;
+
+    // Initialize the Haskell-based tx evaluator
+    let tx_evaluator = ExternalEvaluator::spawn(chain_config_cache).await?;
```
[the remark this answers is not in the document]
ginnun: This is known and handled in another issue/pr: https://github.com/blockfrost/blockfrost-platform/pull/490

crates/testgen/src/testgen.rs
```diff
@@ -0,0 +1,310 @@
...
+/// The name is 'testgen' since it was initially implemented to generate test cases.
+impl Testgen {
+    /// Starts a new child process.
+    pub fn spawn(variant: &str) -> Result<Self, AppError> {
+        let testgen_hs_path = Self::find_testgen_hs().map_err(AppError::Server)?;
+
+        info!(
+            "Using {} as a fallback CBOR error decoder",
+            &testgen_hs_path
+        );
```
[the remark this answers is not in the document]
ginnun: Not a bad idea, no harm to log more info here. Fixed in 359d8b12.

crates/tx_evaluator/src/wrapper.rs
```diff
@@ -0,0 +1,165 @@
...
+/// - failed eval: top-level 'result' field with error json details
+/// - faulty input: top-level 'fault' field with error string
+pub fn wrap_response_v5(resp: TestgenResponse) -> serde_json::Value {
+    let reflection = generate_reflection();
+    match resp {
+        TestgenResponse::Ok(value) => {
+            // The external evaluator returns a v6-like structure, but we need to convert it to v5.
+            let v6_results: Vec<TxEvalResultV6> =
+                serde_json::from_value(value).expect("ExternalEvaluator: invalid v6 result format");
+
```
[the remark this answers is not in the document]
ginnun: Fixed in 359d8b12.

crates/tx_evaluator/src/wrapper.rs
```diff
@@ -0,0 +1,165 @@
...
+};
+
+fn generate_reflection() -> serde_json::Value {
+    json!({"id": uuid::Uuid::new_v4().to_string()})
+}
+
+/// V5 responses does generate three kind of json:
+/// - successfull eval; top-level 'result' field with exec units
+/// - failed eval: top-level 'result' field with error json details
+/// - faulty input: top-level 'fault' field with error string
```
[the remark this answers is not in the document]
ginnun: Fixed in 359d8b12.

crates/tx_evaluator/src/model/api.rs
```diff
@@ -0,0 +1,1121 @@
...
+            },
+        })
+    }
+}
+
+#[derive(Deserialize, Debug)]
+pub struct Value {
+    pub coins: u64,
+    pub assets: Option<HashMap<String, u64>>, // asset name and number. Asset number can be negative when burning assets but this behaviour changed in Conway. Now it can be only PositiveCoin
+}
```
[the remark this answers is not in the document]
ginnun: Fixed in 359d8b12.

crates/platform/src/api/utils/txs/evaluate/utxos.rs
```diff
@@ -1,6 +1,71 @@
...
+            }
+        },
+    }
+}
+
+fn decode_tx_hex(hex_str: &str) -> Result<Vec<u8>, BlockfrostError> {
+    hex::decode(hex_str).map_err(|e| {
+        BlockfrostError::custom_400(format!("invalid hex-encoded transaction CBOR: {e}"))
+    })
 }
```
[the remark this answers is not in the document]
ginnun: Correct, I added base64 support back to these APIs in 8a7184fbe36f3c2282d54f2dbfea23358dcd65a1

crates/common/src/config.rs
```diff
@@ -28,6 +28,7 @@ pub struct Config {
     pub network: Network,
     pub custom_genesis_config: Option<PathBuf>,
     pub data_node: Option<DataNodeConfig>,
+    pub evaluator: Evaluator,
```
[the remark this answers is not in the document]
ginnun: Removed in 29c32398.

crates/common/src/helpers.rs
```diff
@@ -13,7 +14,10 @@ pub fn binary_or_hex_heuristic(xs: &[u8]) -> Vec<u8> {
     let contains_non_hex = xs.iter().any(|&x| !x.is_ascii_hexdigit());
 
     if !even_length || contains_non_hex {
-        xs.to_vec()
+        use base64::{Engine as _, engine::general_purpose};
```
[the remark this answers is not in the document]
ginnun: No need to check this, I reverted my changes in this function and implemented a new one.

michalrus:
BTW, I found this in my notes:

> Re: https://github.com/blockfrost/blockfrost-platform/pull/378#discussion_r2923147312
>
> tx-eval shouldn’t need to know about HTTP errors

Is it still applicable?

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,977 @@
...
+
+fn convert_protocol_param(
+    pp: CurrentProtocolParam,
+    genesis: GenesisConfig,
+) -> Result<MultiEraProtocolParameters, BlockfrostError> {
+    // Create a Conway protocol parameters struct from the ProtocolParam and GenesisConfig
+    let conway_pp: ConwayProtParams = ConwayProtParams {
+        minfee_a: pp.minfee_a.unwrap() as u32,
+        minfee_b: pp.minfee_b.unwrap() as u32,
+        max_block_body_size: pp.max_block_body_size.unwrap() as u32,
```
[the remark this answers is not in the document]
ginnun: I removed `unwrap`s in f1ad42109fac16397d1dab705ad71afc661ec702

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,977 @@
...
+}
+
+fn convert_system_start(
+    sys_start: SystemStart,
+) -> Result<chrono::DateTime<chrono::FixedOffset>, BlockfrostError> {
+    let naive_date = NaiveDate::from_yo_opt(
+        convert_bigint(&sys_start.year)? as i32,
+        sys_start.day_of_year as u32,
+    )
+    .expect("Invalid system start date");
```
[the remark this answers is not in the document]
ginnun: I got rid of `unwrap`s in f1ad42109fac16397d1dab705ad71afc661ec702.

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,268 @@
...
+            utxos: utxos_cbor,
+        };
+
+        let json = serde_json::to_string(&payload).map_err(|err| {
+            BlockfrostError::internal_server_error(format!(
+                "ExternalEvaluator: Failed to serialize payload: {err}"
+            ))
+        })?;
+
+        let response = self.testgen.send(json).await.map_err(|err| {
```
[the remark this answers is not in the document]
ginnun: Sure, added in 7d926630.

ginnun:
> BTW, I found this in my notes:
> 
> > Re: [#378 (comment)](https://github.com/blockfrost/blockfrost-platform/pull/378#discussion_r2923147312)
> > tx-eval shouldn’t need to know about HTTP errors
> 
> Is it still applicable?

Nope, related part (unwrap) to [#378 (comment)](https://github.com/blockfrost/blockfrost-platform/pull/378#discussion_r2923147312) is completely removed.

But I don't remember the context about HTTP errors.

ginnun:
I can make the CI green after https://github.com/blockfrost/testgen-hs/pull/9 and https://github.com/blockfrost/blockfrost-tests/pull/65 are merged and a new `testgen-hs` is released.

michalrus (changes requested):
Note to self for when this PR is close to a merge:

> Let's have a small PR to move `testgen-hs` handling on `main` in accordance to what the `tx-eval` PR did – for better view of conflicts and what _really_ changed

(requesting changes so that it's not lost)

crates/tx_evaluator/src/ogmios5_response.rs
```diff
@@ -0,0 +1,503 @@
...
+            .get("data")
+            .and_then(|d| d["reason"].as_str())
+            .unwrap_or(message),
+    );
+
+    json!({
+        "EvaluationFailure": {
+            name.to_owned(): {
+                "reason": reason
+            }
```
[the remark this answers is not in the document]
ginnun: This compiles and produces the intended key. I verified it via a standalone test.

crates/node/src/chain_config_watch.rs
```diff
@@ -0,0 +1,212 @@
...
+    // Current slot (may be approximate but close enough for scheduling).
+    let elapsed_ms = now_ms.saturating_sub(slot_config.zero_time);
+    let current_slot = slot_config.zero_slot + elapsed_ms / slot_config.slot_length;
+
+    // Slots into the current epoch.
+    let slot_in_epoch = (current_slot - slot_config.zero_slot) % slot_config.epoch_length;
+    let slots_until_next = (slot_config.epoch_length - slot_in_epoch) % slot_config.epoch_length;
+
+    let ms_until_next = slots_until_next * slot_config.slot_length;
+    Duration::from_millis(ms_until_next) + EPOCH_BOUNDARY_BUFFER
```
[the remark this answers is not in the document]
ginnun: Fixed in 111860b3.

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,408 @@
...
+                // Not on-chain. Last entry wins for duplicate TxIns within additional_utxos.
+                additional_deduped.insert(key, (txin, txout));
+            }
+        }
+
+        if !overlapping.is_empty() {
+            return Err(BlockfrostError::custom_400(format!(
+                "Additional UTxO set contains UTxOs that conflict with on-chain state: {}",
+                overlapping.join(", ")
+            )));
```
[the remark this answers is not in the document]
ginnun: Handled in 111860b3.

ginnun:
> Note to self for when this PR is close to a merge:
> 
> > Let's have a small PR to move `testgen-hs` handling on `main` in accordance to what the `tx-eval` PR did – for better view of conflicts and what _really_ changed
> 
> (requesting changes so that it's not lost)

@michalrus This is in the making.

ginnun:
Failing tests are unrelated to this PR.

For evaluate failure, please see https://github.com/blockfrost/blockfrost-tests/pull/73

Cargo.toml
```diff
@@ -67,13 +69,16 @@ nix = { version = "0.30", default-features = false, features = ["signal"] }
...
 num_cpus = "1"
 # FIXME: use a proper Pallas release after they merge <https://github.com/txpipe/pallas/pull/624>:
-pallas-addresses = { git = "https://github.com/txpipe/pallas.git", rev = "ed422bc0d5a09f77617d4882dd4c71089a8bcdb2" }
-pallas-codec = { git = "https://github.com/txpipe/pallas.git", rev = "ed422bc0d5a09f77617d4882dd4c71089a8bcdb2" }
-pallas-crypto = { git = "https://github.com/txpipe/pallas.git", rev = "ed422bc0d5a09f77617d4882dd4c71089a8bcdb2" }
-pallas-hardano = { git = "https://github.com/txpipe/pallas.git", rev = "ed422bc0d5a09f77617d4882dd4c71089a8bcdb2" }
-pallas-network = { git = "https://github.com/txpipe/pallas.git", rev = "ed422bc0d5a09f77617d4882dd4c71089a8bcdb2" }
-pallas-primitives = { git = "https://github.com/txpipe/pallas.git", rev = "ed422bc0d5a09f77617d4882dd4c71089a8bcdb2" }
-pallas-traverse = { git = "https://github.com/txpipe/pallas.git", rev = "ed422bc0d5a09f77617d4882dd4c71089a8bcdb2" }
+pallas-addresses = { git = "https://github.com/blockfrost/pallas.git", rev = "cfa0c633640f5989e44f50ed27efc38992f9ed0d" }
```
michalrus: There has been one more commit since then in your PR – https://github.com/txpipe/pallas/pull/624/changes/9a83c74125738d198010dcb4d61a3595a6d4d771

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,977 @@
...
+        >(&tx_out_cbor)
+        .map_err(|e| {
+            BlockfrostError::internal_server_error(format!("UTxO deserialization failed: {e}"))
+        })?;
+        let tx_out = pallas_primitives::conway::TransactionOutput::<'_>::PostAlonzo(post_alonzo);
+        let multi_era_out = MultiEraOutput::<'_>::Conway(Box::new(Cow::Owned(tx_out)));
+
+        utxos.insert(TxoRef::from(&multi_era_in), EraCbor::from(multi_era_out));
+    }
+    let mut node = node_pool.get().await?;
```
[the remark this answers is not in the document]
michalrus: I don't see this comment here. Maybe more like `FIXME`, not `TODO`? 🫣

crates/integration_tests/tests/data/supported_endpoints.json
```diff
@@ -37,5 +37,7 @@
   "/epochs/latest/parameters",
   "/genesis",
   "/addresses/{address}/transactions",
-  "/addresses/{address}/utxos"
+  "/addresses/{address}/utxos",
+  "/utils/txs/evaluate",
```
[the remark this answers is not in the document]
michalrus: > I should underline, we don't have mempool integration yet (Ogmois does use mempool + chain)

- So the implementation is not equivalent, even though we're deferring to Haskell code? 
- Shouldn’t we add it before releasing? 👀 
- How hard is it?
- And why isn’t it caught by [`blockfrost-tests`](https://github.com/blockfrost/blockfrost-tests)?

crates/node/src/chain_config_watch.rs
```diff
@@ -0,0 +1,233 @@
...
+                        RETRY_INTERVAL.as_secs()
+                    );
+                    tokio::select! {
+                        () = time::sleep(RETRY_INTERVAL) => {},
+                        () = tx.closed() => {
+                            tracing::info!("ChainConfigWatch: stopping monitor");
+                            return;
+                        },
+                    }
+                },
```
michalrus, line 130: So between the epoch boundary and successful refresh, stale protocol params would be used by eval? And the refresh might only practically fail if a node is down?

crates/node/src/chain_config_watch.rs
```diff
@@ -0,0 +1,233 @@
...
+impl ChainConfigWatch {
+    /// Spawn the background monitor and return immediately
+    ///
+    /// The watch value starts as `None` and is set to `Some(…)` once the node
+    /// is synced (tip within one epoch of expected slot)
+    pub fn spawn(node_pool: NodePool) -> Self {
+        let (tx, rx) = watch::channel(None);
+
+        tokio::spawn(async move {
+            monitor_loop(node_pool, tx).await;
```
michalrus: Can this panic? 👀 It seems rather crucial for correct eval results, and we wouldn't restart it then, no?

crates/node/src/chain_config_watch.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// Wait until the first config is available (node synced and init complete)
+    pub async fn wait_ready(&mut self) {
+        while self.rx.borrow().is_none() {
+            match self.rx.changed().await {
+                Ok(_) => {},
+                Err(err) => {
+                    tracing::error!(
+                        "ChainConfigWatch: watch channel closed before configuration became available: {err}"
+                    );
+                    break;
```
michalrus: But then we also return `()` from here, and the caller (`initialize_app` in `crates/integration_tests/src/platform/mod.rs`) would continue? This is only used in tests?

crates/integration_tests/tests/api/utils/txs/evaluate/root.rs
```diff
@@ -0,0 +1,286 @@
...
+    use axum::body::to_bytes;
+    use serde_json::json;
+    use tower::ServiceExt;
+
+    use integration_tests::platform::initialize_app;
+
+    use axum::{body::Body, http::Request};
+    use reqwest::Method;
+
+    #[tokio::test]
```
michalrus: I noticed that the new integration tests are missing `#[ntest::timeout]` like other integration tests, but I don't know if they could hang? The other ones can, and that's why we added this attribute

crates/integration_tests/tests/api/utils/txs/evaluate/root.rs
```diff
@@ -0,0 +1,286 @@
...
+            response.status().is_success(),
+            "Response was not successful"
+        );
+
+        // Convert the response body to bytes
+        let body_bytes = to_bytes(response.into_body(), usize::MAX)
+            .await
+            .expect("Failed to read response body");
+
+        let body_json: serde_json::Value = serde_json::from_slice(&body_bytes).unwrap();
```
michalrus: The tests look pretty repetitive, wouldn't it be more readable to have a function that implements each one? Or even `#[rstest]` and `#[case(…)]` like e.g. here:

https://github.com/blockfrost/blockfrost-platform/blob/3bd7bfde4d1a6273c8cc58c7e43aa8c769567914/crates/platform/src/payment_cred.rs#L150-L158

crates/integration_tests/tests/api/utils/txs/evaluate/utxos.rs
```diff
@@ -0,0 +1,140 @@
...
+    use axum::{
+        body::{Body, to_bytes},
+        http::Request,
+    };
+    use reqwest::Method;
+
+    /// This test is identical to the Blockfrost test:
+    /// https://github.com/blockfrost/blockfrost-tests/blob/dc33312126e9d7c49836d7605ab72224a337bc91/src/fixtures/preview/utils/txs-evaluate-utxos.ts#L5
+    /// https://github.com/blockfrost/blockfrost-tests/blob/dc33312126e9d7c49836d7605ab72224a337bc91/src/fixtures/preview/utils/txs-evaluate-utxos.ts#L22
+    #[tokio::test]
```
michalrus: Same comments – missing `#[ntest::timeout]` + repetitveness (`#[rstest]` maybe?)

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,425 @@
...
+            state: Arc::new(Mutex::new(EvaluatorState {
+                testgen: None,
+                last_config: None,
+            })),
+        }
+    }
+
+    /// Get or init testgen, re-spawning if config changed. Returns 503 if
+    /// chain config is not yet available
+    async fn ensure_testgen(&self) -> Result<Testgen, BlockfrostError> {
```
michalrus: I'm not sure that I like that the `exec()` syscall for starting a subprocess is potentially triggered by each end user request 🤔 A small bug later on, and they can DoS our server?

Why don’t we run and control it in the background, in some Tokio thread that also monitors the output of `ChainConfigWatch`, and here just _talk_ to it? This would seem less DoS-prone

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,425 @@
...
+        // Config unchanged, testgen already running
+        if let (Some(last), Some(testgen)) = (&state.last_config, &state.testgen)
+            && Arc::ptr_eq(last, &current_config)
+        {
+            return Ok(testgen.clone());
+        }
+
+        tracing::info!("ExternalEvaluator: spawning testgen-hs");
+
+        let testgen = spawn_and_init_testgen(&current_config).await.map_err(|e| {
```
michalrus: I missed that in the previous review, but why is it calling `std::thread` instead of spawning a Tokio task?

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,425 @@
...
+            tracing::error!("ExternalEvaluator: failed to initialize testgen-hs: {e}");
+            BlockfrostError::internal_server_error(format!(
+                "Failed to initialize ExternalEvaluator: {e}"
+            ))
+        })?;
+
+        state.testgen = Some(testgen.clone());
```
michalrus: Where is the old process being killed, after a config update from `ChainConfigWatch`?

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,425 @@
...
+
+    /// Evaluates already-decoded CBOR binary. Used by JSON endpoints (utxos.rs).
+    pub async fn evaluate_binary_tx_v5(
+        &self,
+        node_pool: NodePool,
+        tx_cbor_binary: &[u8],
+        additional_utxos: Option<AdditionalUtxoSet>,
+    ) -> Result<serde_json::Value, BlockfrostError> {
+        // Pre-Alonzo tx (3-element CBOR array) cannot be evaluated.
+        if tx_cbor_binary.first() == Some(&0x83) {
```
michalrus: Are you 100% sure that only pre-Alonzo starts with `0x83`? And that no pre-Alonzo can start with something else?

crates/tx_evaluator/src/external.rs
```diff
@@ -0,0 +1,425 @@
...
+
+    /// Evaluates already-decoded CBOR binary. Used by JSON endpoints (utxos.rs).
+    pub async fn evaluate_binary_tx_v6(
+        &self,
+        node_pool: NodePool,
+        tx_cbor_binary: &[u8],
+        additional_utxos: Option<Vec<AdditionalUtxoV6>>,
+    ) -> Result<serde_json::Value, BlockfrostError> {
+        // Pre-Alonzo tx (3-element CBOR array) cannot be evaluated.
+        if tx_cbor_binary.first() == Some(&0x83) {
```
michalrus: Same question

crates/tx_evaluator/src/helper.rs
```diff
@@ -0,0 +1,93 @@
...
+pub fn generate_reflection_v6() -> serde_json::Value {
+    serde_json::Value::String(uuid::Uuid::new_v4().to_string())
+}
+
+/// Decodes a CBOR payload from an `application/cbor` endpoint body.
+/// Matches Blockfrost API behavior:
+/// - Non-ASCII bytes (or empty) → treat as raw binary CBOR
+/// - ASCII text → try hex decode, then base64 decode
+/// - If neither works → error
+pub fn resolve_tx_body(body: &[u8]) -> Result<Vec<u8>, String> {
```
michalrus, line 18: We have this function somewhere else already, no? I remember writing it, hmm

Edit: here:

https://github.com/blockfrost/blockfrost-platform/blob/3bd7bfde4d1a6273c8cc58c7e43aa8c769567914/crates/common/src/helpers.rs#L4-L20

crates/tx_evaluator/src/helper.rs
```diff
@@ -0,0 +1,93 @@
...
+    let is_text = !body.is_empty() && body.iter().all(|b| b.is_ascii());
```
michalrus, line 19: Is empty CBOR valid CBOR?

crates/tx_evaluator/src/native.rs
```diff
@@ -0,0 +1,954 @@
...
+    ))
+}
+
+fn convert_protocol_param(
+    pp: CurrentProtocolParam,
+    genesis: &GenesisConfig,
+) -> Result<MultiEraProtocolParameters, BlockfrostError> {
+    // Create a Conway protocol parameters struct from the ProtocolParam and GenesisConfig
+    let conway_pp: ConwayProtParams = ConwayProtParams {
+        minfee_a: pp.minfee_a.required("minfee_a")? as u32,
```
michalrus: Why are they `u64` in `CurrentProtocolParam`, but `u32` in `ConwayProtParams`? Are these multiple conversions always safe?

crates/tx_evaluator/src/ogmios5_response.rs
```diff
@@ -0,0 +1,503 @@
...
+
+fn transform_utxo_in_error(utxos: Option<Vec<Value>>) -> Option<Vec<Value>> {
+    utxos.map(|utxos| {
+        utxos
+            .into_iter()
+            .map(|utxo| {
+                let tx_id = utxo
+                    .get("transaction")
+                    .and_then(|transaction| transaction.get("id"))
+                    .unwrap_or(&Value::Null);
```
michalrus, line 74: Is the `null` okay here, or should we error?

crates/tx_evaluator/README.md
```diff
@@ -0,0 +1,151 @@
+# TX evaluator
+
+This package contains native and external tx evaluator.
+
+- Native is based on `pallas-validate`.
+- External is using `ledger` codebase via `testgen` Haskell binary.
+
+`native` produces different evaluation results when compared to `external` at the moment. Knowing that `external` depends on the ledger codebase, we can consider `external` results accurate.
```
michalrus, line 8: But we don’t implement the mempool part there? 👀

crates/tx_evaluator/src/wrapper.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// Checks whether the evaluation results represent success by inspecting only the first element.
+/// The evaluator always returns a homogeneous list — either all successes or all failures —
+/// so the first element is sufficient to determine the outcome.
+pub fn is_success_v6(results: &[TxEvalResultV6]) -> bool {
+    if let Some(first_result) = results.first() {
+        matches!(first_result, TxEvalResultV6::Success(_))
+    } else {
+        true
```
michalrus: Can you explain this `true` here? Why are empty `results` a success?

crates/tx_evaluator/src/model/api.rs
```diff
@@ -0,0 +1,1169 @@
...
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(untagged)]
+pub enum EvaluationError {
+    Evaluation(serde_json::Value),
+    Deserialization(DeserializationErrorData),
+}
```
michalrus, line 233: Since this is `#[serde(untagged)]`, and the first variant is arbitrary JSON, the deserializer would never reach the second variant 🤔 Do we ever use the deserializer? If not, maybe let's drop it?

michalrus (commented):
A couple questions – the most important one is about the mempool omission:

crates/tx_evaluator/README.md
ginnun: Yes, no mempool handling. Haskell part doesn't fetch any data from any source. By design, it's the Rust parts that needs to fetch any data and pass it to forward.

crates/node/src/chain_config_watch.rs
ginnun: @michalrus Yes that is correct. Alternatively we can fail (or wait) evaluate calls on epoch boundary but that might be fragile. Another (bad) option is to never cache those values.

crates/node/src/chain_config_watch.rs
ginnun: There is nothing that can panic meaningfully at the moment (only setting slot=0 will panic). But for future code changes that's a possibility, I've added a supervisor loop in 70480e66a7218e2a63c8cda865242c72fbee1f91

crates/tx_evaluator/src/native.rs
ginnun: For `minfee_*` fields:
In pratice this is safe for now. But depending on CDDL(`uint`) and the ledger(`Word64`) code, it should be `u64`. I'm not sure why upstream `pallas` uses `u32`. I changed the `pallas` part in my `pallas` PR since actual data can be bigger. `CurrentProtocolParam` uses correct types based on CDDL.

For other casted fields:
These are completely safe to cast into `u32`. CDDL defines `uint4/8`  and the ledger code uses `Word16/32`. I am not sure again, why `pallas` upstream uses `u64`. Since this won't create a problem ever, I didn't modify `pallas`.

Please see c26479f542ae74bb00e675425fa6482a13673df6

Cargo.toml
ginnun: Updated in c26479f542ae74bb00e675425fa6482a13673df6

crates/tx_evaluator/src/native.rs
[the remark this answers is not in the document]
ginnun: Ah sorry for the confusion. I think I fixed this while I was implementing the `ChainConfigWatch`. Line 79 reads cached values and later they are fed into here.

```Rust
    let config = config_watch.get()?;
```

crates/tx_evaluator/src/ogmios5_response.rs
ginnun: Never gonna happen but if happens, this is ok. I think returning the root error is more meaningful than erroring out because of this unwrap.

crates/integration_tests/tests/data/supported_endpoints.json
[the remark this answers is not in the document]
ginnun: > So the implementation is not equivalent, even though we're deferring to Haskell code?

Yes, but it's not related to Rust or Haskell. It's about what we feed to evaluation logic.
> Shouldn’t we add it before releasing? How hard is it?

When this implementation was completed, pallas didn't have capability to fetch mempool from the node. This is why it is missing. I don't think this should be a blocker.
> And why isn’t it caught by blockfrost-tests?

We don't have a test case that can catch this.

crates/tx_evaluator/src/model/api.rs
ginnun: Ah yes, only `Evaluation` variant is used to/from JSON. But `Deserialization` varian is also used, only to JSON. So both variant is used, but it's confusing. Hmmm.

crates/tx_evaluator/src/wrapper.rs
ginnun: Good catch. The empty case is produced upstream in `external.rs` when the tx has no redeemers (no Plutus scripts to evaluate); we short-circuit testgen-hs and return []. I added clarification in 7d51d8cc972a29762e00342f007b049c4691a96e

crates/integration_tests/tests/api/utils/txs/evaluate/root.rs
ginnun: +1, done in d2bb5c1e178d53cb81bc0ffc000fb3feda4f902b

crates/integration_tests/tests/api/utils/txs/evaluate/utxos.rs
ginnun: +1, done in d2bb5c1e178d53cb81bc0ffc000fb3feda4f902b

crates/integration_tests/tests/api/utils/txs/evaluate/root.rs
ginnun: I did some tidying up in d2bb5c1e178d53cb81bc0ffc000fb3feda4f902b

crates/node/src/chain_config_watch.rs
ginnun: Only for testing, should not harm anything. I did add feature gating since I am here and also changed the returning strategy in cde811b9f474d05c456caa756cfb6580f43b20a8

crates/tx_evaluator/src/external.rs
ginnun: Yes for sure. I checked the CDDLs again.
But for clearness, I changed `== Some(&0x83)` to `!= Some(&0x84)`. Also added a reminder TODO for dijkstra. It will need lots of changes including pallas anyways. You can see in 7a8f22ae655d7387d5c40fba501417c1640da770

crates/tx_evaluator/src/external.rs
ginnun: Yes for sure. I checked the CDDLs again.
But for clearness, I changed `== Some(&0x83)` to `!= Some(&0x84)`. Also added a reminder TODO for dijkstra. It will need lots of changes including pallas anyways. You can see in 7a8f22ae655d7387d5c40fba501417c1640da770

crates/tx_evaluator/src/helper.rs
ginnun: Close but not the same.

This one has `Base64` support and returns `Result` to surface decoding failures. I didn't reuse `binary_or_hex_heuristic` so the `submit` endpoint can keep its infallible contract while `evaluate` surfaces decode errors as 400. Not sure if submit endpoint should have base64 support. Happy to extract a shared inner hex-detection helper (and use it, tho not my liking) or use one implementation with `Base64` support, if you'd prefer.

crates/tx_evaluator/src/helper.rs
ginnun: No, empty isn't valid CBOR. But hosted Blockfrost apis accepts an empty body and surfaces it as an Ogmios deserialisation fault. We match that by passing empty to downstream which produces the same shape of error. Updated the doc to be honest about that in fb7d63d632199358fbc24d27c5db3789ac9bd54c

crates/tx_evaluator/src/external.rs
ginnun: Since this workload is a blocking indefinitely-long loop, I intentionally went with a dedicated thread. Of course it's up to debate, but tokio documentation suggests similar: https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html#when-to-use-spawn_blocking-vs-dedicated-threads

The supervisor task itself is `tokio::spawn` though, and the async caller boundary is `tokio::sync::mpsc`. (see 2df81e6c1148ae66569819bff171ce4e69eae3bc for latest)

crates/tx_evaluator/src/external.rs
ginnun: Moved the (re)spawn loop into a background testgen_supervisor task that only fires on `Arc<ChainConfigCache>` changes. Request path is now a lock-free `watch::Receiver::borrow()` See 2df81e6c

crates/tx_evaluator/src/external.rs
ginnun: Graceful drain. Publishing a new `Testgen` into the watch slot drops the old reference; in-flight request clones keep the worker alive until they finish, then the mpsc receiver disconnects and `testgen::spawn_child` runs `child.kill() + wait()` ([testgen.rs:294](https://github.com/blockfrost/blockfrost-platform/blob/feat/tx-evaluation-rebased/crates/testgen/src/testgen.rs#L294)).

PR #395, opened 2025-09-24
chore: don’t push the `latest` OCI tag for pre-releases

michalrus:
## Context

I noticed that the [`latest` tag points to `0.0.3-rc.3`](https://github.com/blockfrost/blockfrost-platform/pkgs/container/blockfrost-platform), which is a pre-release (release candidate):

Let’s keep it at the latest stable non-RC release?

If we agree, I'll force-push `latest` back to `0.0.2`.


michalrus:
`latest` moved back to `0.0.2`:

image

PR #417, opened 2025-11-24
fix: run all tests in CI

michalrus:
## Context

Previously we set `--all-features` which set the "negative" `tarpaulin` feature, which excluded a great deal of them.

There were even compilation errors hidden under this feature, e.g.:

```
❯ cargo test --package node proptest_ApplyTxErr_Conway_size_010

   Compiling common v0.0.3-rc.3 (/home/mw/Work/blockfrost-platform/crates/common)
error[E0599]: no method named `json` found for struct `reqwest::Response` in the current scope
  --> crates/common/src/json_client.rs:48:25
   |
48 |         let body = resp.json::<T>().await?;
   |                         ^^^^ method not found in `reqwest::Response`
```

## To-do

- [x] the `--lib` property tests are failing outside of the Nix devshell – something wrong with `testgen-hs` download code?

### DIFF
--- a/nix/internal/windows.nix
+++ b/nix/internal/windows.nix
@@ -51,8 +51,10 @@ in rec {
       inherit cargoArtifacts GIT_REVISION;
       doCheck = false; # we run Windows tests on real Windows on GHA
       postPatch = ''
-        sed -r '/^build = .*/d' -i Cargo.toml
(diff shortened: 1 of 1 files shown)


michalrus:
Eh, something with the Windows build still…

Issue #399, opened 2025-10-13
Refactor: only support a single data node


michalrus:
## Context

We went with `--dolos-node` initially, and planned to support multiple nodes explicitly.

Instead, we want `--data-node`, and only a single URL.

If the user wants to run multiple data nodes, they have to route it themselves, e.g. with Nginx, and give the Platform the final (and single) URL.

PR #431, opened 2026-01-14
refactor: single node support

vladimirvolek:
resolves: https://github.com/blockfrost/blockfrost-platform/issues/399

bonus:  human readable cli tests


crates/common/src/cli.rs
```diff
@@ -21,8 +21,24 @@ fn should_skip_serializng_fields<T>(_: &T) -> bool {
...
+        match self {
+            DataNodeType::Dolos => write!(f, "dolos"),
+        }
+    }
+}
+
 #[derive(Parser, Debug, Clone, Serialize, Deserialize, Default)]
-pub struct DolosArgs {
+pub struct DataNodeArgs {
+    pub node_type: DataNodeType,
```
michalrus: I'll submit a more detailed review soon, but why do we need this, if there can only be a single data node?

(IIRC, the idea was that if an SPO wants to mix and match, they should set up their own reverse proxy, like `nginx`, `haproxy`, and still give us only one URL.)

vladimirvolek: It would be nice to have at least in the data node root endpoint, a name or something.

PR #448, opened 2026-02-04
Update all deps + one CI

vladimirvolek:
- CI: merge git-checks into CI, add hadolint, fix duplicate runs

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -35,11 +35,11 @@
     "blockfrost-tests": {
       "flake": false,
       "locked": {
-        "lastModified": 1770733010,
+        "lastModified": 1770896946,
         "narHash": "sha256-4cKMeE/8GptnxWnB6LliJy3gCvjVI34TXTwQHEsHcUA=",
         "owner": "blockfrost",
         "repo": "blockfrost-tests",
-        "rev": "5297592395d1dbd46e88247e459896838c854340",
+        "rev": "79d02c3455b92a9055e2e54b2872f7903c729a15",
         "type": "github"
       },
       "original": {


.github/workflows/oci-image.yaml
```diff
@@ -2,6 +2,8 @@ name: OCI
 
 on:
   push:
+    branches: [main]
+  pull_request:
```
michalrus: This trigger will get an (arguably) wrong `GIT_REVISION=${{ github.sha }}` below, because on the `pull_request:` trigger, GitHub creates a new temporary commit: a merge between the tip of the PR and its target (usually `main`).

Didn’t it behave similarly before this addition? The only change is that now it wouldn’t run for branches that don’t have PRs (yet).

.github/workflows/ci.yaml
```diff
@@ -112,26 +116,26 @@ jobs:
...
-            timeout_minutes: 15
+            timeout: 15
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
 
       - name: Run Blockfrost Tests
         continue-on-error: true
-        timeout-minutes: ${{ fromJSON(matrix.timeout_minutes) }}
+        timeout-minutes: ${{ matrix.timeout }}
```
michalrus: Oh, nice, I thought this was a string in the matrix!

.github/workflows/ci.yaml
```diff
@@ -173,3 +177,110 @@ jobs:
       - name: Check coverage
         timeout-minutes: 10
         run: cargo tarpaulin --workspace --all --lib --features tarpaulin --fail-under 11
+
+  oci_build:
```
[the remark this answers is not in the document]
michalrus: This is the most important comment obviously ↑

How it looks now (no dependencies between OCI and regular builds):

image

.github/workflows/ci.yaml
```diff
@@ -10,26 +13,36 @@ on:
         required: true
         default: true
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
michalrus: This is very nice!

michalrus (commented):
One more comment I missed previously:

PR #473, opened 2026-03-11
fix(tests): get rid of `continue-on-error: true`

michalrus:
## Context

Let's get rid of `continue-on-error: true` because it makes us ignore Dolos regressions, e.g.:
- **81 failing** Mainnet tests on **2026-03-10** – [actions/runs/22881069494/job/66384294330](https://github.com/blockfrost/blockfrost-platform/actions/runs/22881069494/job/66384294330)
- but only **41 failing** Mainnet tests on **2026-03-02** – [actions/runs/22574522434/job/65391623114](https://github.com/blockfrost/blockfrost-platform/actions/runs/22574522434/job/65391623114)

## New idea

1. We already have a file with failing IDs – `crates/platform/tests/data/blacklisted_endpoints.json`
1. Let's add all currently failing IDs there – this PR.
1. Let's run the tests with these IDs blacklisted without `continue-on-error: true` – this PR.
1. But let's add another job – one that **tests only blacklisted IDs** (so it this time takes the IDs to run from the same `blacklisted_endpoints.json` file).
1. This new job **fails** if any of the blacklisted tests **passes** – therefore forcing our developer to remove passing tests from `blacklisted_endpoints.json`
    1. E.g. `nix run .#the-tests 2>&1 | tee tests.log`, and then a simple `grep` if it’s still `0 passed`, if not → ❌

## Log how many ignored tests?

```
❯ jq length crates/platform/tests/data/blacklisted_endpoints.json
84
```

84 ignored endpoints in total (on 3 networks) result in 89 tests ignored on Mainnet – since the mapping is not obvious, maybe we could also write a warning in the yarn test output:

> Ignored: **89 tests**

Or something like this?

### DIFF
--- a/nix/internal/windows.nix
+++ b/nix/internal/windows.nix
@@ -218,8 +218,6 @@ in rec {
       outFileName = "${blockfrost-platform.pname}-${blockfrost-platform.version}-${inputs.self.shortRev or "dirty"}-${targetSystem}.zip";
     } ''
       cp -r ${bundle} ${packageName.pname}
-
-
       mkdir -p $out
       zip -q -r $out/$outFileName ${packageName.pname}/
 
--- a/treefmt.toml
+++ b/treefmt.toml
@@ -1,3 +1,0 @@
-[formatter.nix]
-command = "alejandra"
-includes = ["*.nix"]


nix/internal/unix.nix
```diff
@@ -560,79 +585,115 @@ in
...
+          ''
+          + (
+            if ignorelistOnly
+            then ''
+              set +x
+
+              IGNORELIST_ONLY=true yarn test:${lib.escapeShellArg network} 2>&1 | tee tests.log || true
+
+              # Fail if any ignored test now passes
+              if grep -E 'Tests.*passed' tests.log; then
```
michalrus: This check seems quite wide, e.g. if someone wrote a comment from a test to stdout that:

> Tests that […] passed on 2026-[…]

But then we’d just get a red CI to fix, so I guess it’s fine if it works now!

PR #513, opened 2026-04-07
chore(dolos): update to 1.0.3

michalrus:
## Context

Let's have the latest one in:
- the devshell,
- installers,
- and Docker Compose.

I have also added a CI check to verify that the Dolos configs generated for OCI are the same as the ones we use in the devshell.


ginnun (approved):
lgtm

PR #522, opened 2026-04-10
fix(hydra): reset stale credits on Head `Close` and seed Gateway balance on `Open`

michalrus:
### Issue 1

Resolves:
- https://github.com/blockfrost/blockfrost-platform/actions/runs/24190234830/job/70615404273

After a `hydra-node` crash-restart that re-joins an `Open` head, the Bridge kept stale `credits_available` allowing HTTP 200 responses while `AccountOneRequest` retried forever (Head not `Open`). The Gateway also double-counted the existing snapshot balance as new microtransactions, triggering a spurious Close.

- Bridge: zero `credits_available` in `MonitorStates` when head leaves `Open` and in `on_head_open()` as a defense.

- Gateway: seed `credits_last_balance` from the current snapshot in `WaitForOpen` so only genuinely new deposits are counted.

### Issue 2

Fixes:
- <https://github.com/blockfrost/blockfrost-platform/actions/runs/24266878515/job/70864753840?pr=522>

When both `hydra-node`s crashed, multiple `Event::Restart` events would pile up, each initiating a new KEx round. Responses from older rounds would then race on the gateway's single-session state, causing port mismatches and permanently preventing recovery.

### DIFF
--- a/crates/gateway/src/hydra_server_bridge/mod.rs
+++ b/crates/gateway/src/hydra_server_bridge/mod.rs
@@ -611,8 +611,19 @@ impl State {
                     self.customer_log_id, status
                 );
                 if status == "Open" {
+                    // Seed credits_last_balance from the current snapshot so
+                    // that MonitorCredits does not double-count pre-existing
+                    // funds (e.g. after a hydra-node crash-restart that
+                    // re-joins an already-Open head).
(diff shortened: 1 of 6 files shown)


ginnun (approved):
Rust-wise looks good. I have no idea on the business logic.

PR #528, opened 2026-04-14
fix: immediately run the built binaries; fail early if no `hydra-node`

michalrus:
## Context

- An idea I had – let’s run each binary with `--version` right after it's been built/

- But also fail immediately in `fn main` when the helper binaries are not available.

- Since we treat the helper binaries as "functions", this checks that the entire service is "loaded" correctly.

- One bug was already found – we didn’t add `hydra-node` to the `Dockerfile`

  - … which is a problem, because they're not publishing `aarch64-linux` binaries or images… https://github.com/cardano-scaling/hydra/releases/tag/1.0.0 – and we have multi-arch images.

  - To solve it, I zipped the result of `nix build --builders '' --max-jobs 0 -L .#internal.aarch64-linux.bundle-hydra` locally, and published to https://github.com/blockfrost/hydra-aarch64-linux/releases/tag/1.0.0 – I think it should be good enough…

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ FROM base AS hydra
 ARG TARGETARCH
 # hadolint ignore=DL3008
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends unzip \
+  && apt-get install -y --no-install-recommends unzip patchelf \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
 RUN set -eux; \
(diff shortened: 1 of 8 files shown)


michalrus:
OK, I think that's finally going to work on AArch64 Linux…

Edit: yes! 😮‍💨 – https://github.com/blockfrost/blockfrost-platform/actions/runs/24403504032/job/71280527478?pr=528

michalrus:
@vladimirvolek, CI is green! ✅

vladimirvolek (approved):

PR #532, opened 2026-04-14
fix: hardcoded project_id

### LINKED ISSUE #530: [Maybe] Don’t hardcode secrets in integration tests
## Context

- It's a controversial "secret" because it's only a Preview project ID:

  https://github.com/blockfrost/blockfrost-platform/blob/6484d4f2cf0d09a0094a61038636708400f9ea1e/crates/integration_tests/tests/platform_submit.rs#L72-L76

- Still, it could be considered a secret, and come out during the audit.

- Instead, we could use a GitHub Secret here, and just add it to `.envrc.local` for the devshell.

vladimirvolek:
resolves: https://github.com/blockfrost/blockfrost-platform/issues/530

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -164,6 +164,7 @@ jobs:
         timeout-minutes: 10
         env:
           CARDANO_NODE_SOCKET_PATH: /run/cardano-node/node_preview.socket
+          BLOCKFROST_PREVIEW_PROJECT_ID: ${{ secrets.BLOCKFROST_PREVIEW_PROJECT_ID }}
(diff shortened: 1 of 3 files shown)


michalrus (approved):
Oh, we had two! I missed that, indeed, before:

```
❯ git grep -E '"(preview|preprod|mainnet)[^" .-]' | wc -l
2
```

After:

```
❯ git checkout 8dfeac1ae4dc98b4a15f8dd577ae33b77be39374
HEAD is now at 8dfeac1 fix: hardcoded project_id

❯ git grep -E '"(preview|preprod|mainnet)[^" .-]' | wc -l
0
```

LGTM!

PR #537, opened 2026-04-17
Extract testgen-hs handling into dedicated crates

ginnun:
Extracted from #378 per review feedback, to make the tx-evaluation PR's diff and conflicts easier to inspect.

Moves the `testgen-hs` handling that used to live inside `crates/node/src/cbor/` into two dedicated crates:

- **`bf-testgen`** — the `Testgen` struct that drives the `testgen-hs` binary (variant-aware spawn, child PID tracking, JSON response handling). Moved from `node/src/cbor/fallback_decoder.rs`.
- **`bf-error-decoder`** — the `ExternalDecoder` wrapper plus the random and specific CBOR proptest suites that cross-validate `pallas-hardano` decoding against the Haskell reference. Moved from `node/src/cbor/tests/`.

After the move, `crates/node/src/cbor/` only keeps `validation`, and the `node` crate drops the `crossbeam`, `serde_json`, `sysinfo`, and `num_cpus` dependencies it no longer needs.

Also bumps `testgen-hs` to `10.6.3.1` (latest) and switches to the `blockfrost/testgen-hs` fork (release binaries + Windows URL; Nix flake input follows the same tag).

#378 (`feat/tx-evaluation-rebased`) will be rebased on top of this PR once merged, so the tx-evaluation diff no longer carries the `testgen-hs` move.


nix/internal/windows.nix
```diff
@@ -109,7 +109,7 @@ in rec {
   in
     pkgs.fetchzip {
       name = "testgen-hs-${version}";
-      url = "https://github.com/input-output-hk/testgen-hs/releases/download/${version}/testgen-hs-${version}-${targetSystem}.zip";
+      url = "https://github.com/blockfrost/testgen-hs/releases/download/${version}/testgen-hs-${version}-${targetSystem}.zip";
       hash = "sha256-LXE1RBKgal1Twh7j2hpCfNLsBMEcqSwGHb4bj/Imd9Q=";
```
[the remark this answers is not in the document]
ginnun: This will not be a problem.

flake.nix
```diff
@@ -37,7 +37,7 @@
     };
     mithril.url = "github:input-output-hk/mithril/2524.0";
     testgen-hs = {
-      url = "github:input-output-hk/testgen-hs/10.6.3.0"; # make sure it follows cardano-node
+      url = "github:input-output-hk/testgen-hs/10.6.3.1"; # make sure it follows cardano-node
       flake = false; # otherwise, +2k dependencies we don’t really use
```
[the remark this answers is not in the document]
ginnun: Fixed in 6456f90c3d3afbcd3ccbd50af71a375d393999d1

crates/testgen/src/testgen.rs
```diff
@@ -0,0 +1,369 @@
...
+                .unwrap_or_default(),
+        );
+
+        debug!("{} search paths = {:?}", exe_name, search_paths);
+
+        // Checks if the path is runnable. Adjust for platform specifics if needed.
+        // TODO: check that the --version matches what we expect.
+        fn is_our_executable(path: &Path) -> bool {
+            Command::new(path).arg("--version").output().is_ok()
+        }
```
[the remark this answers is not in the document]
ginnun: Out of scope for this PR, we did not change this code. Recorded in https://github.com/blockfrost/blockfrost-platform/issues/538

crates/testgen/src/testgen.rs
```diff
@@ -0,0 +1,369 @@
...
+        init_payload: Option<&str>,
+    ) -> Result<(), String> {
+        let mut child = proc::Command::new(testgen_hs_path)
+            .arg(variant)
+            .stdin(proc::Stdio::piped())
+            .stdout(proc::Stdio::piped())
+            .spawn()
+            .map_err(|err| format!("couldn’t start the child: {err:?}"))?;
+
+        current_child_pid.store(child.id(), atomic::Ordering::Relaxed);
```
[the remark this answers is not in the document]
ginnun: Out of scope, captured in https://github.com/blockfrost/blockfrost-platform/issues/539

crates/testgen/src/testgen.rs
```diff
@@ -0,0 +1,369 @@
...
+        );
+
+        // Let’s make sure it’s dead in case a different error landed us here.
+        // Will return Ok(()) if already dead.
+        child
+            .kill()
+            .map_err(|err| format!("couldn’t kill the child: {err:?}"))?;
+        child
+            .wait()
+            .map_err(|err| format!("couldn’t reap the child: {err:?}"))?;
```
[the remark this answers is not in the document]
ginnun: Out of scope, captured in https://github.com/blockfrost/blockfrost-platform/issues/539

crates/node/src/cbor.rs
```diff
@@ -1,5 +1 @@
-pub mod fallback_decoder;
-
-#[cfg(test)]
-pub mod tests;
 pub(crate) mod validation;
```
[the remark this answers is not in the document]
ginnun: Not a stale-code issue. These files are moved, not left behind.

nix/internal/windows.nix
```diff
@@ -109,7 +109,7 @@ in rec {
...
     };
```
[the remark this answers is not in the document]
ginnun: Not valid.

crates/error_decoder/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    bf_build_utils::testgen_hs::ensure();
```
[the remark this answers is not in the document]
ginnun: This is an already existing behavior, I see no urgency to change.

crates/testgen/src/testgen.rs
```diff
@@ -0,0 +1,369 @@
...
+                };
+
+                let _ = request.response.send(response);
+            }
+
+            // Now break the loop, and restart everything if we failed:
+            result_for_logs?
+        }
+
+        Err("request channel closed".to_string())
```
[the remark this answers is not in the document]
ginnun: Misread. `?` on `Ok(())` doesn't return; it unwraps and the loop continues. Only `Err` breaks out.

crates/testgen/src/testgen.rs
```diff
@@ -0,0 +1,369 @@
...
+                Ok(Self::Ok(json))
+            },
+            (Some(json), None) => Ok(Self::Ok(json)),
+            (None, Some(error)) => Ok(Self::Err(error)),
+            (None, None) => Err(de::Error::custom(
+                "invalid testgen-hs response: missing both `json` and `error`",
+            )),
+        }
+    }
+}
```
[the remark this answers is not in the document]
ginnun: Added tests in 405d4f78587c5a901f27e4a4aef19601aa5238e9

Cargo.toml
```diff
@@ -12,6 +12,8 @@ members = [
   "crates/api_provider",
   "crates/node",
   "crates/data_node",
+  "crates/testgen",
+  "crates/error_decoder",
```
michalrus, line 16: Quick question before I read the entire PR – can you elaborate why we need separate crates? 🙏 It's not clear from the PR description

ginnun: `testgen` hosts the base & shared parts of the external binary communication. `error_decoder` hosts the relevant logic to error decoding while depending on `testgen`. tx evaluation will do the same (another pr).

michalrus: Yes, yes, that's clear, but I guess I meant more like: why separate crates, and not just separate modules? Something like that

ginnun: This way looked cleaner to me, it gives better separation.

crates/error_decoder/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[dependencies]
+bf-common.workspace = true
+bf-testgen.workspace = true
+hex.workspace = true
+sysinfo.workspace = true
+serde_json.workspace = true
+tokio.workspace = true
+pallas-network.workspace = true
+pallas-hardano.workspace = true
+pallas-codec.workspace = true
```
michalrus: `sysinfo`, `pallas-network`, `pallas-hardano`, `pallas-codec`, and `tokio` are only used in `#[cfg(test)]` code. They should be `[dev-dependencies]`, not `[dependencies]`. I tested this move, and everything still compiles!

crates/testgen/src/testgen.rs
```diff
@@ -0,0 +1,445 @@
...
+impl Testgen {
+    /// Starts a new child process.
+    pub fn spawn(variant: Variant) -> Result<Self, AppError> {
+        Self::spawn_inner(variant, None)
+    }
+
+    /// Starts a new child process with an init payload that is sent as the first
+    /// message on every (re)start. The response to the init payload is validated
+    /// but not returned — it is only used to confirm the subprocess is ready.
+    pub fn spawn_with_init(variant: Variant, init_payload: String) -> Result<Self, AppError> {
```
michalrus, line 100: Dead code?

crates/testgen/src/tests.rs
```diff
@@ -1,18 +1,9 @@
 #[cfg(not(feature = "tarpaulin"))]
```
michalrus: I wonder how this works, if at the usage site, `num_cpus::get` is not guarded behind a similar `#[cfg(…)]` pragma:

https://github.com/blockfrost/blockfrost-platform/blob/bc68d97be495a03ac0d32781f8cb660388b05a71/crates/testgen/src/tests.rs#L78-L82

crates/testgen/src/testgen.rs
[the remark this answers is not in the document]
michalrus: Wait, why out of scope? Isn't this more/less a one-liner? Something like:

```rust
        let _ = child.kill();
        let _ = child.wait();
        current_child_pid.store(u32::MAX, atomic::Ordering::Relaxed);
```

crates/testgen/src/testgen.rs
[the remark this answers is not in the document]
michalrus: The same comment as above:

Wait, why out of scope? Isn't this more/less a one-liner? Something like:

```rust
        let _ = child.kill();
        let _ = child.wait();
        current_child_pid.store(u32::MAX, atomic::Ordering::Relaxed);
```

crates/testgen/src/testgen.rs
```diff
@@ -0,0 +1,445 @@
...
+            let payload = request.payload.clone();
+            *last_unfulfilled_request = Some(request);
+
+            let mut ask_and_receive = || -> Result<Result<TestgenResponse, String>, String> {
+                writeln!(stdin, "{payload}")
+                    .map_err(|err| format!("couldn’t write to stdin: {err:?}"))?;
+
+                match stdout_lines.next() {
+                    Some(Ok(line)) => Ok(Ok(serde_json::from_str::<TestgenResponse>(&line)
+                        .map_err(|e| e.to_string())?)),
```
michalrus, line 355: This new `?` here – won’t a JSON parse error kill and restart the child process? Are we fine with this?

michalrus (commented):
Okay, a couple of comments.

In my original review of #378, I just meant some uncontroversial _moving around_ of existing files, so that their future #378 locations match on `main`, without introducing many changes to them. This would allow us to see the actual changes in the other PR.

Right now, to see what really changed in the most important file, I still have to do something like:

```
diff -uw <(git show origin/main:crates/node/src/cbor/fallback_decoder.rs) ./crates/testgen/src/testgen.rs
```

And there are quite a few changes in it:

```
1 file changed, 299 insertions(+), 146 deletions(-)
```

A regular `git diff` (either local or in GitHub) just shows it as a new file, so we never see the actual changes to logic that were introduced, and that triggered my initial comment in #378. 😀

I just mean that I also had to run this extra command in #378.

Anyhow, now it's split, so #378 will be a little easier to review.

BTW,

> #378 (`feat/tx-evaluation-rebased`) will be rebased on top of this PR once merged, so the tx-evaluation diff no longer carries the `testgen-hs` move.

A rebase of that colossal #378 sounds like a real lot of work, maybe a merge? 🫣

Cargo.toml
michalrus: Modules also provide separation, I was just curious why 😇

ginnun: The logical grouping I've chosen is what functionality a piece of code is providing, instead of how they are implemented underneath. To me, a `crate` is more natural in this case. For example, `evaluator` crate (in another PR) contains external and native implementation in one crate, but as separate modules.

crates/testgen/src/testgen.rs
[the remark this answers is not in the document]
ginnun: @michalrus For the sake of keeping it slim, I am trying to limited scope of this PR to moving things around and preparing for `evaluator` implementation. It's not about the size of the change.

I'll happily fix this recorded issue in a separate PR. Nothing is lost.

crates/testgen/src/testgen.rs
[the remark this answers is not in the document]
ginnun: Same as above:

@michalrus For the sake of keeping it slim, I am trying to limited scope of this PR to moving things around and preparing for evaluator implementation. It's not about the size of the change.

I'll happily fix this recorded issue in a separate PR. Nothing is lost.

crates/testgen/src/testgen.rs
ginnun: This function will be used in `evaluator` crate. Since this PR is taken out of a bigger PR, it was much easier for me to split the final version of `testgen` crate.

crates/testgen/src/tests.rs
ginnun: Good catch! This works because we are using absolute path as num_cpus::get() and import is dropped by the compiler since it's a no-op.

Removed in 02c4a397fb0e503358b7aa613e2cb3a936b760f3

ginnun: Interesting bit is that our CI checks not cacthing this.

crates/error_decoder/Cargo.toml
ginnun: Yeah, removed in 02c4a397fb0e503358b7aa613e2cb3a936b760f3

crates/testgen/src/testgen.rs
ginnun: I think this is the right call. A malformed JSON on stdout might mean the child's output stream is desynchronized. we can't reliably recover by reading the next line, since we don't know whether it belongs to the failed request or the next one. Failing fast is the safest option.

Note we don't immediately fail the user-facing request. The retry logic (`is_a_retry`) gives us one second attempt against a fresh subprocess, and only on the second failure do we return `"repeated internal failure"` to the caller. So a transient glitch (e.g., a  corrupted line for whatever reason) won't surface to the user.

ginnun:
> In my original review of https://github.com/blockfrost/blockfrost-platform/pull/378, I just meant some uncontroversial moving around of existing files, so that their future https://github.com/blockfrost/blockfrost-platform/pull/378 locations match on main, without introducing many changes to them. This would allow us to see the actual changes in the other PR.

Separating an existing complex PR was already a lot of work. To do it in the cleanest way possible required much more work. This is the midpoint I've preferred. This version should be much simpler to review, despite the need for extra commands.



> A rebase of that colossal https://github.com/blockfrost/blockfrost-platform/pull/378 sounds like a real lot of work, maybe a merge? 🫣

I've split in a way that they don't conflict much. Let's see when the time comes, but I am always in favor of rebasing if could be done at a fair cost.

crates/testgen/src/testgen.rs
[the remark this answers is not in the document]
michalrus: It's also like that on the current main, right?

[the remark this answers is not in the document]
michalrus: Yes:

https://github.com/blockfrost/blockfrost-platform/blob/e150fbc79810784e0638cd1533bb9225409f6480/crates/node/src/cbor/fallback_decoder.rs#L146-L155

crates/testgen/src/testgen.rs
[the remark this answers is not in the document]
michalrus: It's also like that on current `main`:

https://github.com/blockfrost/blockfrost-platform/blob/e150fbc79810784e0638cd1533bb9225409f6480/crates/node/src/cbor/fallback_decoder.rs#L146-L155

Cargo.toml
michalrus: Hmm, I dug a little and the consensus seems to be:
- Use _modules_ for organization, visibility, and local flexibility.
- Use _crates_ when you want a real dependency/**compilation boundary**, especially one that improves reuse, build parallelism, or architectural separation.
  - I.e. do not split into crates just because "more modular" sounds better, the dependency graph has to buy you compilation time.

Based on:
- https://doc.rust-lang.org/book/ch07-00-managing-growing-projects-with-packages-crates-and-modules.html
- https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html
- https://doc.rust-lang.org/reference/crates-and-source-files.html
- https://mmapped.blog/posts/03-rust-packages-crates-modules.html
- https://matklad.github.io/2021/09/04/fast-rust-builds.html
- https://matklad.github.io/2021/08/22/large-rust-workspaces.html

crates/testgen/src/testgen.rs
michalrus: I see, hmmmm. But it's three lines of code (not counting comments) 😀

crates/testgen/src/tests.rs
michalrus: > Interesting bit is that our CI checks not cacthing this.

Yes, hmmm. Unused imports 🤷 Maybe it's some corner case in the compiler

crates/testgen/src/testgen.rs
michalrus: Fair enough!

michalrus (commented):
My main reservation is that the `blockfrost-tests` are failing!

I'm not 100% sure about the crate/module split, given the sources above, but I don’t care that much.

You think we could fix the tests, also on `main`? I'll look into it. @vladimirvolek also has a branch where he updates them, I'll check if they're passing there.

crates/testgen/src/testgen.rs
ginnun: @michalrus I should admit that the PR separation isn't the cleanest one.

michalrus:
@ginnun let's merge the latest `main` into this, and then we can merge this, once CI passes :pray:

michalrus:
There are also some Git conflicts in `flake.lock`

michalrus:
> @ginnun let's merge the latest `main` into this, and then we can merge this, once CI passes :pray: 

I've just pushed it myself, because it was quite trivial, and CI takes a while

crates/testgen/src/testgen.rs
```diff
@@ -289,13 +289,13 @@ impl Testgen {
             variant,
         );
 
-        // Make sure the child is dead before returning.
-        child
+        let _ = child
             .kill()
-            .map_err(|err| format!("couldn’t kill the child: {err:?}"))?;
-        child
+            .inspect_err(|err| debug!(err = %err, "testgen-hs: child pid kill failed"));
```
michalrus: I wonder if this is a `debug!` or more like `warn!`?

michalrus (approved):
LGTM – thanks!

PR #548, opened 2026-04-22
refactor: common

### LINKED ISSUE #518: Refactor: `crates/common/` is only Platform-common, not Gateway–
## Context

- E.g. Platform’s CLI shouldn’t be in `crates/common/` which is shared with Gateway etc. Isn’t a more proper place for CLI just somewhere near `main.rs`?

- What else is there that is not shared between Platform–Gateway–Bridge?

- Possibly CLI-related: #480

vladimirvolek:
resolves: https://github.com/blockfrost/blockfrost-platform/issues/518


michalrus:
There are some Git conflicts:
- `Cargo.lock`
- `crates/common/Cargo.toml`
- `crates/common/src/lib.rs`
- `crates/platform/Cargo.toml`
- `crates/platform/src/main.rs`

crates/common/src/lib.rs
```diff
@@ -1,24 +1,9 @@
...
-pub mod cli;
-pub mod config;
-pub mod crates;
-pub mod dreps;
-pub mod epochs;
 pub mod errors;
 pub mod find_libexec;
-pub mod genesis;
 pub mod hydra;
 pub mod json_client;
```
michalrus, line 5: Only used by `crates/data_node/`, we could consider moving it there.

crates/common/src/lib.rs
```diff
@@ -1,24 +1,9 @@
...
-pub mod metadata;
 pub mod pagination;
```
michalrus, line 6: Only used by `crates/platform/`, likewise.

michalrus (approved):
Two rather minor comments, otherwise, LGTM, so let's merge. Thanks!

PR #551, opened 2026-04-24
feat: add new endpoints

### LINKED ISSUE #456: Re-think `ci.yaml` dependencies between jobs
Follow-up to:
* https://github.com/blockfrost/blockfrost-platform/pull/448#discussion_r2793601769

## Context

After we have a single `ci.yaml` file, it’s worth thinking about dependencies – this is what that workflow merge in #448 really gives us.

Do we want to:
* save GitHub minutes, and not pay for OCI builds that would fail because we saw that `build_and_test`,
* or have better developer turn-around time, but running as much as possible in parallel, even if all the builds would fail because of a single compilation error?

The same question about Hydra tests:
* maybe it makes sense to run them only if the PR is fine?
* OTOH, that will increase the total wall clock "build" time by 15 minutes, etc.

vladimirvolek:
### Added

- `/pools/{pool_id}`
- `/pools/{pool_id}/metadata`
- `/accounts/{stake_address}/withdrawals`

also resolves: https://github.com/blockfrost/blockfrost-platform/issues/456

image


ginnun (approved):

crates/integration_tests/tests/data/supported_endpoints.json
```diff
@@ -14,9 +14,12 @@
   "/network",
   "/accounts/{stake_address}/delegations",
   "/accounts/{stake_address}/registrations",
+  "/accounts/{stake_address}/withdrawals",
   "/accounts/{stake_address}",
   "/assets/{asset}",
   "/pools/extended",
+  "/pools/{pool_id}",
+  "/pools/{pool_id}/metadata",
```
michalrus, line 22: Looking good – 🚩 we need to wait for a new Dolos for this to work 📝

michalrus:
It looks like all 3 `blockfrost-tests` runs are failing 👀

There are also some Git conflicts

PR #563, opened 2026-05-11
ci: smart cache and install + more in parallel

vladimirvolek:
Related to:
- #456

~19 minutes -> ~8 minutes

image


.github/workflows/ci.yaml
```diff
@@ -159,7 +154,7 @@ jobs:
   platform_integration_tests:
     name: Platform Integration Tests
     runs-on: [self-hosted, Linux, X64, nixos]
-    needs: [build_and_test, check_flake_lock]
+    needs: [check_flake_lock]
```
michalrus, line 157: Nice, this `runs-on: self-hosted`, and will fail fast (if there's a problem) most of the time, because our `runner1` has Cargo deps cached in the Nix build! I.e. no need to wait for a non-self-hosted build

michalrus (approved):
Very nice!

PR #577, opened 2026-05-27
feat: translations

vladimirvolek:


michalrus:
The `treefmt` checks are failing with some Markdown white space nits, could you run `nix fmt` in the root of the repository? :pray:

Or just `treefmt` while inside the devshell.

PR #587, opened 2026-06-15
Fix  fetching custom network file

vladimirvolek:


michalrus:
Something happened to the commit message of 91032235546848cdceddfcbdda1e30e58f2178f5 😅:

michalrus (commented):
BTW, after this work, do you have any thoughts about sync progress:

https://github.com/blockfrost/blockfrost-platform/blob/b82dbb06df1d0eed2ffe279621a35b6b3afc0e57/crates/node/src/sync_progress.rs#L24-L35

michalrus (commented):
I also found this – so according to us, no stake address is valid on any custom network:

https://github.com/blockfrost/blockfrost-platform/blob/b82dbb06df1d0eed2ffe279621a35b6b3afc0e57/crates/platform/src/addresses.rs#L74-L87

vladimirvolek:
> I also found this – so according to us, no stake address is valid on any custom network:
> 
> https://github.com/blockfrost/blockfrost-platform/blob/b82dbb06df1d0eed2ffe279621a35b6b3afc0e57/crates/platform/src/addresses.rs#L74-L87

wow very good catch!

crates/platform/src/config.rs
```diff
@@ -136,39 +147,70 @@ impl Config {
...
+        ("update_quorum", genesis.update_quorum),
+    ];
+
+    for (field, value) in non_negative {
+        if value < 0 {
+            return Err(format!("`{field}` must be non-negative, got {value}"));
+        }
     }
+
+    Ok(())
```
michalrus, line 233: Maybe let's also check:
- `active_slots_coefficient` is an `f64` between `(0,1]`
- `max_lovelace_supply` is a positive integer

michalrus: Maybe we should also reject magic numbers that shadow existing and well-known networks? So that user's `GenesisRegistryMut::add` with a well-known number doesn't break searching networks by magic… OTOH, one run of `blockfrost-platform` is bound to a specific network, so maybe it doesn’t matter that much. Like _when_ would it matter?

crates/platform/src/genesis.rs
```diff
@@ -17,12 +17,14 @@ pub trait GenesisRegistry {
 
     /// Map a magic number back to its `Network`.
     fn network_by_magic(&self, magic: u64) -> &Network;
+}
 
-    /// Insert or replace the `GenesisContent` for `network` at the front.
+pub trait GenesisRegistryMut {
```
michalrus, line 18: Now `pub type GenesisExtension` above appears unused:

https://github.com/blockfrost/blockfrost-platform/blob/b82dbb06df1d0eed2ffe279621a35b6b3afc0e57/crates/platform/src/genesis.rs#L6

crates/platform/src/config.rs
```diff
@@ -27,6 +27,7 @@ pub struct Config {
     pub no_metrics: bool,
     pub network: Network,
     pub custom_genesis_config: Option<PathBuf>,
+    pub genesis: Vec<(Network, GenesisResponse)>,
```
michalrus, line 31: Okay, so now this is redundant:

https://github.com/blockfrost/blockfrost-platform/blob/b82dbb06df1d0eed2ffe279621a35b6b3afc0e57/crates/platform/src/server/state.rs#L8-L13

Maybe let's have all handlers reference a single source of truth: the new `state.config.genesis` (the line I'm commenting this on).

Removing that previous one will also surface places that are still using it (if any).

michalrus (commented):
A couple more comments:

crates/platform/src/config.rs
michalrus: > Maybe let's also check:

OTOH… The user wouldn't be able to run a `cardano-node` otherwise 🤔 I don’t know, your call

michalrus:
A few Git conflicts appeared after recent merges :octocat:

michalrus:
@vladimirvolek, is this ready for a re-review?

crates/platform/src/config.rs
```diff
@@ -138,39 +149,88 @@ impl Config {
...
+
+    validate_genesis(&custom).map_err(|e| {
+        AppError::Server(format!(
+            "Invalid custom genesis file {}: {}",
+            path.display(),
+            e
+        ))
+    })?;
+
+    Ok(Some(custom))
```
michalrus, line 190: I wonder, maybe we should reject known magics (1 = Preprod, 2 = Preview, 764824073 = Mainnet)?

Otherwise the user can overwrite them… But it's their fault… So maybe it's OK.

crates/platform/src/config.rs
```diff
@@ -28,6 +28,7 @@ pub struct Config {
     pub no_metrics: bool,
     pub network: Network,
     pub custom_genesis_config: Option<PathBuf>,
```
michalrus, line 30: According to my LSP's `lsp-find-references`, this is never used, now that the genesis is parsed up front into `genesis`.

crates/platform/src/config.rs
```diff
@@ -138,39 +149,88 @@ impl Config {
...
+    for (field, value) in non_negative {
+        if value < 0 {
+            return Err(format!("`{field}` must be non-negative, got {value}"));
         }
+    }
 
-        Ok(registry)
+    let asc = genesis.active_slots_coefficient;
+
+    if !(asc > 0.0 && asc <= 1.0) {
```
michalrus, line 217: Is this written in this form instead of `asc <= 0.0 || asc > 1.0` to also detect NaN? But implicitly. 😀 Clever, but maybe an explicit `asc.is_nan()`. Someone might "simplify" and introduce a bug.

Anyway, it's user's responsibility to provide correct genesis, so we're probably fine with that hypothetical bug in the future.

crates/platform/src/genesis.rs
```diff
@@ -47,7 +45,9 @@ impl GenesisRegistry for Vec<(Network, GenesisResponse)> {
             .map(|(n, _)| n)
             .expect("Missing Network for known magic")
     }
+}
 
+impl GenesisRegistryMut for Vec<(Network, GenesisResponse)> {
     fn add(&mut self, network: Network, genesis: GenesisResponse) {
         // If the network already exists, replace its GenesisContent;
         // otherwise, insert this (network, genesis) tuple at index 0.
```
michalrus, line 53: Similar comment to the first one, inserting at index 0, makes an overridden magic win the network-by-magic lookups. Probably fine as-is, it's power user's responsibility if they shoot themselves (themself? 🙄) in the foot.

michalrus (approved):
LGTM! 4 comments which you're probably fine to ignore 😀

PR #598, opened 2026-06-30
feat: add /blocks/slot/{slot_number}

vladimirvolek:


docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
       init-node:
         condition: service_completed_successfully
   dolos:
-    image: ghcr.io/txpipe/dolos:v1.3.2
+    image: ghcr.io/txpipe/dolos:v1.2.0
```
michalrus: The newest one is 1.3.2 -- it's already on `main`

crates/platform/src/server/routes/hidden.rs
```diff
@@ -45,6 +45,7 @@ pub fn get_hidden_api_routes(enable_metrics: bool) -> Router<AppState> {
 
         // blocks
         .route("/blocks/epoch/{epoch_number}/slot/{slot_number}", get(blocks::epoch::epoch_number::slot::slot_number::route))
+        .route("/blocks/slot/{slot_number}", get(blocks::slot::slot_number::route))
```
michalrus, line 48: Should we enable `blockfrost-tests` for this one?

docker-compose.yml
michalrus: (Hence the Git conflict)

michalrus: Fixed in 00843630992b239119d3ef81e01af820fec9e084

crates/platform/src/server/routes/hidden.rs
michalrus: Fixed in 8fb22e61a21a0b214eb49a2983c4ef675ab682bc.

I think this is a bug in `blockfrost-tests`:
- `"/blocks/slot/{slot}"` was being silently ignored,
- `"/blocks/slot/{slot_number}"` works.

We probably should get a big error if someone adds something that doesn't make sense to `supported_endpoints.json` -- otherwise how do we catch it?

michalrus (approved):
LGTM ✅

crates/platform/src/server/routes/hidden.rs
michalrus: Said bug is (hopefully) fixed in:
- https://github.com/blockfrost/blockfrost-tests/pull/95

PR #606, opened 2026-07-13
feat: more gateway metrics

### LINKED ISSUE #602: Gateway: expose per-Platform `blockfrost_gateway_relay_healthy`, `_has_data_node`, and `_version`
## Context

Follow-up to:
- #553 
- #589

In Gateway metrics, let's monitor each relay’s:
- health,
- version,
- whether it has a data node connected.

Non-200 response should be consider unhealthy (the Platform will return `http/503` if it's unhealthy!).

Both these facts can be see in each Platform’s `GET /` response. We can periodically call that from within the Gateway, for all connected Platforms, without changing the WebSocket protocol.

So this is a Gateway-side-only change.

Maybe this leads to better design: it would also be nice to have these 3 data points under `GET /stats`, not only in metrics.

vladimirvolek:
resolves: https://github.com/blockfrost/blockfrost-platform/issues/602


hlolli:
@copilot review the prometheus metric names if they follow the prometheus style/naming conventions described here https://prometheus.io/docs/practices/naming/ and if they don't please suggest better name

hlolli:
@copilot turn this into inline review on the places with a suggestion block

hlolli:
This is not what I asked Copilot to do 😩. Please verify @vladimirvolek if this commit it did was ok, otherwise revert. Looks ok to me from a glance.

hlolli (approved):

crates/gateway/src/load_balancer.rs
```diff
@@ -713,12 +759,18 @@ pub mod event_loop {
                 },
 
                 LBEvent::NewRelayMessage(RelayMessage::Response(response)) => {
+                    let code = response.code;
+                    let is_health_check =
+                        pass_on_response(response, &relay_state, asset_name).await;
+                    // Only bill known user requests to Hydra micropayments —
+                    // neither our own health checks, nor unknown (e.g. already
+                    // expired and cleaned-up) requests:
```
michalrus, line 767: So one other change this introduces is that the end user no longer micro-pays for a timed out request.

Since the relay timed out, maybe the work was significant?

But I think this is fair to the user, they don't care, if they didn't get a response.

crates/gateway/src/load_balancer.rs
```diff
@@ -1186,6 +1246,74 @@ pub mod event_loop {
...
+
+/// A background task to periodically check the relay’s platform over the
+/// regular websocket
+async fn check_platform_health_periodically(
+    new_request_channel: mpsc::Sender<RequestState>,
+    platform_health: Arc<Mutex<Option<PlatformHealth>>>,
+) {
+    let mut interval = tokio::time::interval(HEALTH_CHECK_INTERVAL);
+    loop {
+        interval.tick().await;
```
michalrus, line 1263: The first `tick` returns immediately, so we ask Platform for health right after it connects to the WebSocket.

I wonder if it won't result in a brief unhealthiness.

OTOH, if we didn't, we'd show `null`s there, which is also unhealthy, hmm.

OK, let's see how this works in practice!

crates/gateway/src/load_balancer.rs
```diff
@@ -1186,6 +1246,74 @@ pub mod event_loop {
...
+        let health = check_platform_health(&new_request_channel).await;
+        *platform_health.lock().await = Some(health);
+    }
+}
+
+/// Sends a single `GET /` to the relay’s platform, exactly like a regular
+/// proxied HTTP request, i.e. without changing the WebSocket protocol.
+async fn check_platform_health(new_request_channel: &mpsc::Sender<RequestState>) -> PlatformHealth {
+    tokio::time::timeout(HEALTH_CHECK_TIMEOUT, async {
```
michalrus, line 1272: Hmmm, I don't think you need this `timeout` since you set `expires:` below. 

It _could_ in principle happen that you reach this timeout a few ns before `now() + HEALTH_CHECK_TIMEOUT`, and then the request will only get cleaned up by the 60 s cleanup task, which logs a "request timed out" warning. Noisy but harmless. 🤷

I also don't think that Platform's `GET /` would often take more than 15 seconds, so it doesn't really matter. 😀

michalrus (approved):
Great, LGTM! ✨

PR #610, opened 2026-07-15
feat: gateway health monitoring

### LINKED ISSUE #601: Gateway: `GET /` shouldn’t always return `healthy: true`
## Context

It shouldn’t always return `healthy: true`, especially when there's a connection issue in the DB pool.

We had this as a production issue, and it went unnoticed for 6 days, until @opacular and @hlolli found it, see:

- https://github.com/blockfrost/blockfrost-platform/pull/599
- https://github.com/blockfrost/blockfrost-ops/issues/2887

We should also monitor non-200 responses to `POST /register`, which were also failing during said outage, and we didn't notice them, because they're not instrumented anywhere.

Another failure mode on 2026-07-07 was `blockfrost-gateway` getting `http/402` Payment Required from the Blockfrost API, when running `POST /register`.

And here's the offending code:

https://github.com/blockfrost/blockfrost-platform/blob/a906514081ea105b31698df1690bee3bf3ea1027/crates/gateway/src/api/root.rs#L14-L19

vladimirvolek:
resolves: https://github.com/blockfrost/blockfrost-platform/issues/601

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ proptest-regressions/
 # Nix
 result
 result-*
-/generated
+generated
 
 # direnv
 .direnv/
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ proptest-regressions/
 # Nix
 result
 result-*
-generated
+/generated
 
 # direnv
 .direnv/
--- a/crates/common/src/tracing.rs
+++ b/crates/common/src/tracing.rs
@@ -153,10 +153,14 @@ pub fn setup_tracing(log_level: Level, log_target_env: &str) {
 }
 
 fn setup_default(log_level: Level) {
+    // Human-readable local time instead of the default
+    let timer = tracing_subscriber::fmt::time::ChronoLocal::new("%Y-%m-%d %H:%M:%S%.3f".to_owned());
+
     tracing_subscriber::fmt()
         .with_max_level(log_level)
         .event_format(
             Format::default()
+                .with_timer(timer)
                 .with_ansi(true)
                 .with_level(true)
                 .with_target(true)
--- a/crates/common/src/tracing.rs
+++ b/crates/common/src/tracing.rs
@@ -153,14 +153,10 @@ pub fn setup_tracing(log_level: Level, log_target_env: &str) {
 }
 
 fn setup_default(log_level: Level) {
-    // Human-readable local time instead of the default
-    let timer = tracing_subscriber::fmt::time::ChronoLocal::new("%Y-%m-%d %H:%M:%S%.3f".to_owned());
-
     tracing_subscriber::fmt()
         .with_max_level(log_level)
         .event_format(
             Format::default()
-                .with_timer(timer)
                 .with_ansi(true)
                 .with_level(true)
                 .with_target(true)
--- a/crates/gateway/CHANGELOG.md
+++ b/crates/gateway/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### Fixed
 
 - `GET /` no longer always reports `healthy: true` — it now reflects periodic health checks of PostgreSQL connectivity and the Blockfrost API
+- The Gateway now refuses to start when the initial health check fails (e.g. an invalid Blockfrost project token)
 - The underlying Blockfrost API error (e.g. rate limiting) is now logged when the license NFT check fails during registration
 
 ### Removed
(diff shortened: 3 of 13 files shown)


crates/common/src/tracing.rs
```diff
@@ -153,10 +153,14 @@ pub fn setup_tracing(log_level: Level, log_target_env: &str) {
 }
 
 fn setup_default(log_level: Level) {
+    // Human-readable local time instead of the default
+    let timer = tracing_subscriber::fmt::time::ChronoLocal::new("%Y-%m-%d %H:%M:%S%.3f".to_owned());
```
michalrus: I'm not sure if it's better than the explicit timezone in the original RFC-3339 / ISO-8601 format, but we already do the same in `crates/platform`

crates/gateway/src/main.rs
```diff
@@ -42,6 +42,16 @@ async fn main() -> Result<()> {
...
     let blockfrost_api = blockfrost::BlockfrostAPI::new(&config.blockfrost.project_id);
+    let health_monitor =
+        health_monitor::HealthMonitor::spawn(pool.clone(), blockfrost_api.clone()).await;
+
+    // Fail fast on startup problems
+    let initial_health = health_monitor.current_status().await;
+    if !initial_health.healthy {
+        tracing::error!("Refusing to start unhealthy: {}", initial_health.details());
+        std::process::exit(1);
+    }
```
michalrus, line 53: So a transient connectivity problem with the Blockfrost API will abort startup, hmmm. If it happened during operation of this binary, it wouldn't kill it. Slight inconsistency, but it's probably okay, since we run it as a systemd service, so it will be restarted soon after such failure.

.gitignore
```diff
@@ -23,7 +23,7 @@ proptest-regressions/
 # Nix
 result
 result-*
-/generated
+generated
```
michalrus: Why? 👀

crates/gateway/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Added
 
 - Per-relay `blockfrost_gateway_relay_healthy`, `blockfrost_gateway_relay_data_node_up`, and `blockfrost_gateway_relay_info` metrics in `GET /metrics` (and the same data points in `GET /stats`)
+- `blockfrost_gateway_healthy` metric in `GET /metrics`, mirroring the `healthy` field of `GET /`
```
michalrus, line 6: I wonder if we need separate `CHANGELOG`s if we more/less duplicate the gateway entries between them.

crates/gateway/src/health_monitor.rs
```diff
@@ -0,0 +1,129 @@
...
+
+                first_check_done_.notify_one();
+
+                let delay = if healthy {
+                    Self::HEALTHY_CHECK_INTERVAL
+                } else {
+                    Self::UNHEALTHY_CHECK_INTERVAL
+                };
+
+                time::sleep(delay).await;
```
michalrus: Maybe `tokio::time::interval`, which accounts for the time elapsed during work? I really liked this in your previous PR:

https://github.com/blockfrost/blockfrost-platform/blob/288f218f9d2a8b6b296851bc6799e2d54d92e419/crates/gateway/src/load_balancer.rs#L1255-L1261

Say we reach the `DB_PING_TIMEOUT` of 5 s, it would probably better to retry immediately (after 5 seconds), not every 7 seconds etc.

crates/gateway/src/health_monitor.rs
```diff
@@ -0,0 +1,129 @@
...
+                .flatten()
+                .collect();
+                let healthy = errors.is_empty();
+
+                let codes: Vec<&'static str> = errors.iter().map(|e| e.code).collect();
+                let status_ = HealthStatus { healthy, errors };
+
+                if previous_codes.is_some() && previous_codes.as_ref() != Some(&codes) {
+                    if healthy {
+                        tracing::warn!("Gateway became healthy again.");
```
michalrus: Sefa suggested this, but I'm still not sure if it's correct. Becoming healthy again is a warning? Hmmm, maybe, if someone filters by `≥ warning`?

michalrus (commented):
A couple of comments:

crates/gateway/CHANGELOG.md
michalrus: Logged as:
- #620

crates/common/src/tracing.rs
vladimirvolek: It's out of scope od this PR anyway. Removing..

.gitignore
vladimirvolek: When you run `nix develop` in some other folder different then `/`

vladimirvolek: also removed

michalrus (approved):
LGTM -- thanks!

PR #611, opened 2026-07-15
tests: data node smoke tests

### LINKED ISSUE #476: Dolos smoke test – query for real recent data (from legacy API), changing on each run
## Context

- Our current data node tests run https://github.com/blockfrost/blockfrost-tests/

- Maybe let's add something more dynamic, like: get recent transactions from the "legacy" API, and query for them?

- This way we’d be able to catch that it "stopped working".

vladimirvolek:
resolves: https://github.com/blockfrost/blockfrost-platform/issues/476


crates/integration_tests/src/platform/mod.rs
```diff
@@ -109,6 +109,20 @@ pub fn test_config_with_data_node(
...
+/// Serves the app on a random local port and returns its base URL.
+pub async fn spawn_app(app: Router) -> String {
+    let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let addr = listener.local_addr().unwrap();
+
+    tokio::spawn(async move {
+        axum::serve(listener, app).await.unwrap();
+    });
+
+    tokio::task::yield_now().await;
```
michalrus: Why is this needed? 👀

nix/internal/unix.nix
```diff
@@ -715,12 +715,12 @@ in
...
-            if [[ -z ''${DOLOS_ENDPOINT+x} ]]; then
-              export DOLOS_ENDPOINT="http://127.0.0.1:3010"
-              echo >&2 "warning: DOLOS_ENDPOINT is unset; assuming $DOLOS_ENDPOINT"
+            if [[ -z ''${DATA_NODE_ENDPOINT+x} ]]; then
+              export DATA_NODE_ENDPOINT="http://127.0.0.1:3010"
+              echo >&2 "warning: DATA_NODE_ENDPOINT is unset; assuming $DATA_NODE_ENDPOINT"
             fi
 
-            curl -fsSL "''${DOLOS_ENDPOINT}" | jq -r '"Running Dolos " + .version + " (" + .revision + ")"'
+            curl -fsSL "''${DATA_NODE_ENDPOINT}" | jq -r '"Running data node " + .version + " (" + .revision + ")"'
```
michalrus, line 723: A pity there's no `.name` in `GET /`

crates/integration_tests/tests/data_node_smoke.rs
```diff
@@ -0,0 +1,106 @@
...
+    let (app, _, _, _, _) = build_app_with_data_node(data_node_endpoint())
+        .await
+        .expect("Failed to build the application");
+
+    get_platform_client(&spawn_app(app).await)
+}
+
+/// Hash of an already-settled block, so no compared field is racing the chain
+/// tip.
+async fn anchor_block_hash(platform: &BlockfrostAPI) -> String {
```
michalrus: I understand why you chose a local platform (because that block is already available locally), though if the real API is the reference, then I don't know if we shouldn't formally chose it as anchor. Maybe a slightly older block?

michalrus (approved):
LGTM, one minor comment, that's safe to ignore:

crates/integration_tests/src/platform/mod.rs
vladimirvolek: It isn't!

PR #612, opened 2026-07-15
fix: remove api prefix

### LINKED ISSUE #259: Move `GET /metrics` and `GET /` to a different port before we have WebSockets
## Context

Will be superseded by:
* #178 

@mmahut writes:

> I actually think we should move [`GET /`, `GET /metrics`] to a different port, as we do not want these to open to the public (which until we have the websockets will be).
>
> That way people will be able to monitor it without exposing it.

vladimirvolek:
resolves: https://github.com/blockfrost/blockfrost-platform/issues/259


michalrus (commented):
Without reading the code yet -- is it backwards compatible? I.e. the old `crates/platform` (1.0.0) still works with this `crates/gateway`?

michalrus:
> Without reading the code yet -- is it backwards compatible? I.e. the old `crates/platform` (1.0.0) still works with this `crates/gateway`?

/cc @vladimirvolek

PR #622, opened 2026-07-28
chore(blockfrost-tests): unignore now passing Mainnet fixtures

michalrus:
Otherwise, CI on `main` is broken:
- https://github.com/blockfrost/blockfrost-platform/actions/runs/30365275545/job/90294847346


michalrus:
I'm going to merge without reviews, because this exact change has already been reviewed and merged as part of:
- #626

